PR #4, opened 2016-04-27
Fix the condition of retryable errors

jmuk:
Currently it retries if the error doesn't have 'code' attribute,
but that's not right, should raise errors for unexpected failures.


geigerj:
LGTM

Issue #5, opened 2016-04-27
API objects should accept an optional 'client_config' argument


jmuk:
See https://github.com/googleapis/gax-python/issues/95, the same discussion can be applied to Ruby.

jmuk:
Forgotten to mark as fixed

PR #13, opened 2016-05-16
Update construct_settings for config_overrides

jmuk:
For issue #5, this allows overriding configs in the same format
as the default config.


geigerj:
Note: since this is functionally the same as googleaps/gax-python#103, I'll only make language-independent comments over there.

jmuk:
committing to this PR as this behavior is already discussed in googleapis/gax-python#103

PR #19, opened 2016-05-23
Fix some wierd styles

### LINKED ISSUE #15: review and remove remaining Pythonic styles
I got some feedback saying gax-ruby contains a few things which is kinda Pythonic (that's because originally this was based on googleapis/gax-python).  Likewise, underscores-functions or some comments mention things like 'None'.

I'll have to review the entire code and rewrite these things to Ruby style.

jmuk:
This happened because gax-ruby originally borrowed the logic
and documentation string from gax-python -- some things are
not very ruby-ish.

- remove underscore-prefixed names: this is simply Pythonic
  style but not necessary in Ruby. They are already marked
  as private in the code.
- fix comments, mostly s/None/nil/

This fixes #15


tbetbetbe:
LGTM

Issue #18, opened 2016-05-21
Add page_token into CallOptions (and retire is_page_streaming)


jmuk:
From https://github.com/googleapis/toolkit/issues/139:

- CallOptions should accept 'page_token' field as a parameter to start a stream from a specific position
- PagedEnumerable should provide the interface to access to the current response object and the next page token in the current response (they are accessible through `page` instance, but a bit obscure).
- remove is_page_streaming from CallOptions. page-streaming methods should always return PagedEnumerable.

jmuk:
#20 actually fixed this.

PR #20, opened 2016-05-23
Update page-streaming call options

jmuk:
- Retire is_page_streaming call options, we don't use this anymore.
- Introduce page_token as a call option. This is similar to python
  implementation.
- Unlike Python, Ruby provides per-page iteration as with another
  method 'each_page'. I believe this is still good for Ruby users.
- Added example comments and a few fields for better usability
  of PagedEnumerable. I hope this helps.
- related with #18.

### DIFF
--- a/lib/google/gax/api_callable.rb
+++ b/lib/google/gax/api_callable.rb
@@ -136,7 +136,7 @@ module Google
 
       # Initiate the streaming with the requests and keywords.
       # @param page_token [Object]
-      #   The initial page_token or nil.
+      #   The page token for the first page to be streamed, or nil.
       # @param request [Object]
       #   The initial request object.
       # @param kwargs [Hash]
@@ -300,6 +300,8 @@ module Google
     # @param response_page_token_field [String] The field of the next
     #   page token in the response.
     # @param resource_field [String] The field to be streamed.
+    # @param page_token [Object] The page_token for the first page to be
+    #   streamed, or nil.
     # @return [Proc] A proc that returns an iterable over the specified field.
     def page_streamable(a_func,
                         request_page_token_field,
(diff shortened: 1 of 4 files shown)


jmuk:
cc: @tbetbetbe

lib/google/gax/api_callable.rb
```diff
@@ -112,14 +135,17 @@ def initialize(a_func, request_page_token_field,
       end
 
       # Initiate the streaming with the requests and keywords.
+      # @param page_token [Object]
+      #   The initial page_token or nil.
```
geigerj: I think "initial page_token" could be misinterpreted to mean "the page token for the initial page". (E.g., Python has a constant `INITIAL_PAGE`, which means the page token for the initial page overall.)

How about, "The page_token for the for the first page to be streamed, or nil"?

lib/google/gax/api_callable.rb
```diff
@@ -269,12 +304,15 @@ def bundleable(a_func, desc, bundler)
     def page_streamable(a_func,
                         request_page_token_field,
                         response_page_token_field,
-                        resource_field)
+                        resource_field,
+                        page_token)
```
geigerj: Add `page_token` param to doc

geigerj:
Two small nits about the comments, otherwise LGTM.

What is the "codecov/patch" that is failing?

lib/google/gax/api_callable.rb
jmuk: done

lib/google/gax/api_callable.rb
jmuk: done

geigerj:
LGTM

jmuk:
I don't know hwy codecov looks broken -- the link itself is broken and I believe that's codecov's problem.
The report https://codecov.io/gh/googleapis/gax-ruby/commit/d09e902f6e238011d765911441f1e58c0847ff93 seems fine to me. The "85%" means that 85% in the diff is covered by the tests, and that's mostly because of adding several accessors which are not used by tests.

PR #21, opened 2016-05-24
Update handling of errors.

### LINKED ISSUE #16: Fix the use of errors
Feedback points out some mistakes on the use of errors, especially:
- should raise an error object inheriting StandardError, not a string
- GaxError overrides 'cause' field which is defined in Exception class (http://docs.ruby-lang.org/en/2.2.0/Exception.html#method-i-cause)
- some conditional matches on rescue is redundant (https://github.com/googleapis/gax-ruby/blob/master/lib/google/gax/api_callable.rb#L226)

jmuk:
The core part of this patch is to remove 'cause' attribute
from GaxError. Actually Ruby Exception already has such method
from 2.1.0, and $ERROR_INFO contains the previous exception
(or cause) for an exception.

As the consequence of this redesign, we will have to raise the
RetryError inside of rescue clause (otherwise $ERROR_INFO is
not set), and therefore retryable is now an infinite loop.

Also removed an unnecessary if-clause for catching errors,
instead bringing them into the matcher of rescue clause.

And fix some other lint errors:
- raise Error.new(msg) => raise Error, msg -- recommended
  style from rubocop.
- avoid using raise msg -- from internal feedback
- fix some typos

fixes #16


lib/google/gax/errors.rb
```diff
@@ -27,20 +27,26 @@
...
+        msg = "GaxError #{msg}"
+        msg += ", caused by #{$ERROR_INFO}" if $ERROR_INFO
         super(msg)
-        @cause = cause
+        @cause = $ERROR_INFO
+      end
+
+      # cause is a new method introduced in 2.1.0, bring this
+      # method if it does not exist.
+      unless respond_to?(:cause)
```
geigerj: Just to check -- what versions of Ruby are we committed to supporting? It doesn't look like our CI checks with older versions of Ruby, so such code is non-testable at the moment.

If we want to support older versions of Ruby, then we should be testing them with Travis; if we don't want to support them, then we don't need this code.

geigerj:
Nice, the automatic `cause` method seems like a pretty cool feature!

lib/google/gax/errors.rb
jmuk: The oldest supported Ruby version is 2.0.0 -- because gcloud-ruby has this (https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/gcloud.gemspec#L19).

Modified .travis.yml to test with those old Rubies.

geigerj:
LGTM

Issue #14, opened 2016-05-18
Simplify generated API method implementations


garrettjonesgoogle:
Do the analogous simplification described in https://github.com/googleapis/gax-python/issues/104 for Ruby.

PR #22, opened 2016-06-09
Redesign api_callable

jmuk:
This is for #14 and googleapis/gax-python#104.


geigerj:
Do you want to update this to be similar to Python, or leave as-is?

jmuk:
I'll withdraw this PR and return with the approach similar to Python/NodeJS. Sorry for the confusion.

PR #30, opened 2016-07-12
Replace timeout by deadline.

jmuk:
I've noticed that timeout parameter is gone in 0.15.0 -- grpc assumes
'deadline' instead. Since deadline is in grpc-0.14.1, better to
stick with deadline argument.

### DIFF
--- a/spec/google/gax/api_callable_spec.rb
+++ b/spec/google/gax/api_callable_spec.rb
@@ -48,19 +48,19 @@ describe Google::Gax do
   describe 'create_api_call' do
     it 'calls api call' do
       settings = CallSettings.new
-      timeout_arg = nil
-      func = proc do |timeout: nil|
-        timeout_arg = timeout
+      deadline_arg = nil
+      func = proc do |deadline: nil|
+        deadline_arg = deadline
         42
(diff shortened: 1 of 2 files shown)


spec/google/gax/api_callable_spec.rb
```diff
@@ -211,8 +211,8 @@ def initialize(msg, code)
       call_count = 0
 
       time_now = Time.now
-      allow(Time).to receive(:now).exactly(4).times.and_return(
-        *([time_now] * to_attempt + [time_now + 2])
+      allow(Time).to receive(:now).exactly(7).times.and_return(
```
geigerj: What does this change do?

jmuk: This code mocks Time.now, and the number of invocation changes because Time.now can be called twice per API call (one in retryable and the other, the new one, in add_timeout_arg).

Added comments to describe, also replaced the number (7) by the calculation from to_attempt.
PTAL.

jmuk:
@geigerj, can you take a look?

geigerj:
LGTM

PR #31, opened 2016-07-13
Rename delay_threshold by delay_threshold_millis.

jmuk:
The existing code fails with actual bundling configurations,
because construct_settings wants to set the field named as
delay_threshold_millis.

This _millis suffix also matches the style with BackoffSettings,
thus this is probably better.


geigerj:
LGTM

PR #33, opened 2016-07-15
Added bundling to gax-ruby.

Landrito:
This is to get feedback on the initial implementation of bundling in ruby. Do not merge until the template in toolkit is ready and all of the commented out sections pretaining to bundling in gax-ruby are uncommented out.

#6

### DIFF
--- a/lib/google/gax/bundling.rb
+++ b/lib/google/gax/bundling.rb
@@ -289,6 +289,8 @@ module Google
           run_now(bundle.bundle_id)
         end
 
+        # TODO: Implement byte and element count limits.
+
         event
       end
 
--- a/lib/google/gax/bundling.rb
+++ b/lib/google/gax/bundling.rb
@@ -272,8 +272,6 @@ module Google
       # @param bundling_request [Object] the request to pass as the arg to
       #     the api_call.
       # @return [Event] an Event that can be used to wait on the response.
-      #
-      # TODO: Implement byte and element count limits.
       def schedule(api_call, bundle_id, bundle_desc,
                    bundling_request)
         bundle = bundle_for(api_call, bundle_id, bundle_desc,
--- a/lib/google/gax/bundling.rb
+++ b/lib/google/gax/bundling.rb
@@ -31,6 +31,8 @@ module Google
   module Gax
     MILLIS_PER_SECOND = 1000.0
 
+    # TODO: Add function to compute bundle id.
+
     # rubocop:disable Metrics/ClassLength
 
     # Coordinates the execution of a single bundle.
@@ -98,7 +100,7 @@ module Google
       def run
         return if in_deque.count == 0
         request = @bundling_request
-        request[bundled_field] = @in_deque.flatten
+        request.send("#{@bundling_field}=", @in_deque.flatten)
 
         if @subresponse_field
           run_with_subresponses(request, @subreponse_field, @kwargs)
@@ -130,12 +132,13 @@ module Google
         response = @api_call.call(request, **kwargs)
         in_sizes_sum = 0
         in_deque.each { |elts| in_sizes_sum += elts.count }
-        all_subresponses = response[subresponse_field]
+        all_subresponses = response.send(subresponse_field.to_s)
         if all_subresponses.count != in_sizes_sum
           "cannot demultiplex the bundled response, got
             #{all_subresonses.count} subresponses; want #{in_sizes_sum},
             each bundled request will receive all responses"
         end
+        # TODO: handle each subresponse
         @event_deque.each do |event|
           event.result = response
           event.set
@@ -247,7 +250,7 @@ module Google
                    bundling_request, kwargs: {})
         bundle = bundle_for(api_call, bundle_id, bundle_desc,
                             bundling_request, kwargs)
-        elts = bundling_request[bundle_desc.bundled_field]
+        elts = bundling_request.send(bundle_desc.bundled_field.to_s)
         event = bundle.extend(elts)
 
         count_threshold = @options.element_count_threshold
--- a/lib/google/gax/bundling.rb
+++ b/lib/google/gax/bundling.rb
@@ -28,6 +28,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 module Google
+  # Gax defines Google API extensions
   module Gax
     DEMUX_WARNING = [
       'Warning: cannot demultiplex the bundled response, got ',
@@ -267,8 +268,6 @@ module Google
 
     # Organizes bundling for an api service that requires it.
     class Executor
-      UPDATE_THREADS_RATE = 60
-
       # @param [BundleOptions]configures strategy this instance
       #     uses when executing bundled functions.
       def initialize(options)
--- a/lib/google/gax/bundling.rb
+++ b/lib/google/gax/bundling.rb
@@ -275,19 +275,6 @@ module Google
         # Keep track of the threads that are spawned in order to ensure the
         # api call is made before the main thread dies.
         @threads = {}
-
-        # The class may spawn many threads. This thread will remove any dead
-        # threads from @threads.
-        @threads_monitor = Thread.new do
-          loop do
-            sleep UPDATE_THREADS_RATE
-            @tasks_lock.synchronize do
-              @threads.each do |bundle_id, thread|
-                @threads.delete(bundle_id) unless thread.alive?
-              end
-            end
-          end
-        end
       end
 
       # Schedules bundle_desc of bundling_request as part of
@@ -321,6 +308,19 @@ module Google
         event
       end
 
+      # Helper function for #schedule.
+      #
+      # Given a return the corresponding bundle for a certain bundle id. Create
+      # a new bundle if the bundle does not exist yet.
+      #
+      # @param api_call [Proc] used to make an api call when the task is run.
+      # @param bundle_id [String] the id of this bundle.
+      # @param bundle_desc [BundleDescriptor] describes the structure of the
+      #     bundled call.
+      # @param bundling_request [Object] the request to pass as the arg to
+      #     the api_call.
+      # @param kwargs [Hash] the keyword arguments passed to the api_call.
+      # @return [Task] the bundle containing the +api_call+.
       def bundle_for(api_call, bundle_id, bundle_desc, bundling_request, kwargs)
         @tasks_lock.synchronize do
           return @tasks[bundle_id] if @tasks.key?(bundle_id)
@@ -328,21 +328,43 @@ module Google
                             bundling_request, kwargs,
                             bundle_desc.subresponse_field)
           delay_threshold = @options.delay_threshold
-          run_later(bundle, delay_threshold) if delay_threshold > 0
+          run_later(bundle.bundle_id, delay_threshold) if delay_threshold > 0
           @tasks[bundle_id] = bundle
           return bundle
         end
       end
 
-      def run_later(bundle, delay_threshold)
+      # Helper function for #schedule.
+      #
+      # Creates a new thread that will execute the encapsulated api calls after
+      # the +delay_threshold+ has elapsed. The thread that is
+      # spawned is added to the @threads hash to ensure that the thread will
+      # api call is made before the main thread exits.
+      #
+      # @param bundle_id [String] the id corresponding to the bundle that
+      #     is run.
+      # @param delay_threshold [Numeric] the number of micro-seconds to wait
+      #     before running the bundle.
+      def run_later(bundle_id, delay_threshold)
         @tasks_lock.synchronize do
-          Thread.new do
+          thread = Thread.new do
             sleep(delay_threshold / MILLIS_PER_SECOND)
-            run_now(bundle.bundle_id)
+            run_now(bundle_id)
+            @tasks_lock.synchronize do
+              @threads.delete(bundle_id)
+            end
+          end
+          @tasks_lock.synchronize do
+            @threads[bundle_id] = thread
           end
         end
       end
 
+      # Helper function for #schedule.
+      #
+      # Immediately runs the bundle corresponding to the bundle id.
+      # @param bundle_id [String] the id corresponding to the bundle that
+      #     is run.
       def run_now(bundle_id)
         @tasks_lock.synchronize do
           if @tasks.key?(bundle_id)
@@ -352,6 +374,8 @@ module Google
         end
       end
 
+      # This function should be called before the main thread exits in order to
+      # ensure that all api calls are made.
       def close
         @threads.each do |bundle_id, _|
           @tasks_lock.synchronize do
--- a/lib/google/gax/bundling.rb
+++ b/lib/google/gax/bundling.rb
@@ -43,10 +43,11 @@ module Google
     # @param obj [Object] an object.
     # @param name [String] a name for a field in the object.
     # @raise [NoMethodError] if the object does not contain the named attribute.
-    # @return [Object] value of named attribute. Can be nil.
+    # @return [String, nil] value of named attribute. Can be nil.
     def str_dotted_access(obj, name)
       name.split('.').each do |part|
-        obj = obj.send(part)
+        obj = obj[part]
+        break if obj.nil?
       end
       obj.nil? ? nil : obj.to_s
     end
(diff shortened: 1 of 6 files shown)


lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,447 @@
...
+            response_copy = response.dup
+            subresponses = all_subresponses[start, i.count]
+            response_copy.send(@subresponse_field).clear
+            response_copy.send(@subresponse_field).concat(subresponses)
+            start += i.count
+            event.result = response_copy
+            event.set
+          end
+        end
+      rescue Exception => err
```
Landrito: Should these errors be wrapped in GaxError like in api_callable?

spec/google/gax/bundling_spec.rb
```diff
@@ -0,0 +1,656 @@
...
+            expect(event.set?).to eq(true)
+            expect(event.result).to eq(
+              bundled_builder([an_elt] * elts_for_threshold)
+            )
+          end
+        end
+      end
+
+      it 'api call not invoked until time threshold' do
+        asleep = true
```
Landrito: Would there be a better way to do this test? I was having a hard time mocking time and async calls.

spec/google/gax/bundling_spec.rb
```diff
@@ -0,0 +1,661 @@
...
+
+        # This needs to be any instance because the instance of Kernal inside
+        # the thread will not be the main thread's instance of kernal
+        allow_any_instance_of(Kernel).to receive(:sleep) do |_|
```
Landrito: Using 'allow_any_instance_of' on the Kernel is dangerous and can cause unpredictable behavior as seen in this builds travis test failing. I am going to make the test less robust until I find a better way to mock time accross threads.

Landrito:
Integration has been committed @jmuk.

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,456 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+module Google
+  # Gax defines Google API extensions
+  module Gax
+    DEMUX_WARNING = [
+      'Warning: cannot demultiplex the bundled response, got ',
+      '%d subresponses; want %d, each bundled request will ',
+      'receive all responses'
+    ].join
```
jmuk: `private_constant :DEMUX_WARNING` ?

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,456 @@
...
+    # Used to retrieve a nested field signified by name where dots in name
+    # indicate nested objects.
+    #
+    # @param obj [Object] an object.
+    # @param name [String] a name for a field in the object.
+    # @raise [NoMethodError] if the object does not contain the named attribute.
+    # @return [Object] value of named attribute. Can be nil.
+    def str_dotted_access(obj, name)
+      name.split('.').each do |part|
+        obj = obj.send(part)
```
jmuk: This will cause an exception if `obj` becomes `nil` in the middle.  `break if obj.nil?` might be necessary.

Also, Ruby protobuf accepts `[]`, so `obj = obj[part]` might be better (this would be helpful for testing, you can simply use a Hash for the data).

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,456 @@
...
+    ].join
+
+    # Helper function for #compute_bundle_id.
+    # Used to retrieve a nested field signified by name where dots in name
+    # indicate nested objects.
+    #
+    # @param obj [Object] an object.
+    # @param name [String] a name for a field in the object.
+    # @raise [NoMethodError] if the object does not contain the named attribute.
+    # @return [Object] value of named attribute. Can be nil.
```
jmuk: `String, nil` might be the right return type.

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,456 @@
...
+        result.push(str_dotted_access(obj, field))
+      end
+      result
+    end
+
+    # rubocop:disable Metrics/ClassLength
+
+    # Coordinates the execution of a single bundle.
+    #
+    # @attribute [r] bundle_id
```
jmuk: Strictly, `@!attribute` is the right format (see http://www.rubydoc.info/gems/yard/file/docs/Tags.md#attribute).

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,456 @@
...
+        @in_deque = []
+        @event_deque = []
+      end
+
+      # The number of bundled elements in the repeated field.
+      # @return [Numeric]
+      def element_count
+        sum = 0
+        @in_deque.each { |elts| sum += elts.count }
+        sum
```
jmuk: FYI: you could write these three lines as `@in_deque.reduce(0) { |sum, elts| sum + elts.count }`

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,456 @@
...
+                     bundle_id,
+                     bundled_field,
+                     bundling_request,
+                     kwargs,
+                     subresponse_field: nil)
+        @api_call = api_call
+        @bundle_id = bundle_id
+        @bundled_field = bundled_field
+        @bundling_request = bundling_request
+        @kwargs = kwargs
```
jmuk: We actually don't have to have `kwargs`. As the result of recent redesign, kwargs are already passed to api_call at a different place and bundling doesn't have to care about it (see https://github.com/googleapis/gax-ruby/blob/master/lib/google/gax/api_callable.rb#L243).

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,456 @@
...
+            event.result = response
+            event.set
+          end
+        else
+          start = 0
+          @in_deque.zip(@event_deque).each do |i, event|
+            response_copy = response.dup
+            subresponses = all_subresponses[start, i.count]
+            response_copy.send(@subresponse_field).clear
+            response_copy.send(@subresponse_field).concat(subresponses)
```
jmuk: Similarly, this can simply be `response_copy[@subresponse_field] = subresponses`

lib/google/gax/bundling.rb
jmuk: I think normal API Calls are already wrapped by GaxError through catch_errors function (see https://github.com/googleapis/gax-ruby/blob/master/lib/google/gax/api_callable.rb#L260).  So `rescue GaxError` might be just acceptable.

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,456 @@
...
+            false
+          else
+            true
+          end
+        end
+      end
+
+      # private_class_method :run_with_no_subresponse,
+      #                      :run_with_subresponses,
+      #                      :event_for, :cancellor_for
```
jmuk: Remove them if they are actually unnecessary

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,456 @@
...
+    end
+
+    # Organizes bundling for an api service that requires it.
+    class Executor
+      # @param [BundleOptions]configures strategy this instance
```
jmuk: nit: options is missing (`@param options [BundleOptions] configures strategy`...)

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,456 @@
...
+      #     uses when executing bundled functions.
```
jmuk: Please write the comment for timer

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,456 @@
...
+
+        count_threshold = @options.element_count_threshold
+        if count_threshold > 0 && bundle.element_count >= count_threshold
+          run_now(bundle.bundle_id)
+        end
+
+        size_threshold = @options.request_byte_threshold
+        if size_threshold > 0 && bundle.request_bytesize >= size_threshold
+          run_now(bundle.bundle_id)
+        end
```
jmuk: Recommend to write a TODO comment for 'limits'.

For the background, when I was doing bundling for gax-nodejs, @geigerj pointed out that bundling option also has the parameter for 'limit', which GAX Python didn't implement yet. See: https://github.com/googleapis/gax-ruby/blob/master/lib/google/gax.rb#L264

However, from my experience of implementing it for NodeJS, supporting limit is actually more complex than it seems, so I recommend not doing it in this PR -- it's worth doing in another one (it also requires more test cases).  But it's better to leave a TODO comment.

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,456 @@
...
+      end
+
+      # Invokes the cancellation function provided.
+      def cancel
+        @mutex.synchronize do
+          canceller.nil? ? false : canceller.call
+        end
+      end
+
+      def wait(timeout_millis: nil)
```
jmuk: Better to use a condition variable (http://docs.ruby-lang.org/en/2.2.0/ConditionVariable.html) instead of writing a wait loop.

lib/google/gax/bundling.rb
Landrito: Since protobuf needs types to be enforced this will not work. Probably response_copy[@subresponse_field] = []
response_copy[@subresponse_field] += subresponses
https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#repeated-fields

spec/google/gax/bundling_spec.rb
```diff
@@ -0,0 +1,666 @@
...
+      end
+    end
+  end
+
+  describe Google::Gax::Executor do
+    SIMPLE_DESCRIPTOR = Google::Gax::BundleDescriptor.new('field1', [])
+    DEMUX_DESCRIPTOR =
+      Google::Gax::BundleDescriptor.new(
+        'field1', [], subresponse_field: 'field1'
+      )
```
jmuk: For safety, it's better to have an after hook to run `clear` method.

spec/google/gax/bundling_spec.rb
```diff
@@ -0,0 +1,666 @@
...
+          event = bundler.schedule(
+            api_call,
+            bundle_id,
+            DEMUX_DESCRIPTOR,
+            bundled_builder(["#{an_elt}#{i}"] * i)
+          )
+          events.push(event)
+        end
+        previous_event = nil
+        events.each_index do |i|
```
jmuk: It would be more handy with `events.each_with_index do |current_event, i| ...`

jmuk:
By the way, remove 'DO NOT MERGE' from the PR title 😉

lib/google/gax/bundling.rb
jmuk: I see. I also locally tried a bit and noticed that `= []` would be bad.  So keep it the existing 'clear' and 'concat' as
```
response_copy[@subresponse_field].clear
response_copy[@subresponse_field].concat(subresponses)
```

lib/google/gax/bundling.rb
Landrito: Fixed

lib/google/gax/bundling.rb
Landrito: Fixed

lib/google/gax/bundling.rb
Landrito: Fixed

spec/google/gax/bundling_spec.rb
Landrito: Fixed.

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,448 @@
...
+      # Resets the Event to not being set and clears the @result.
+      def clear
+        @mutex.synchronize do
+          @result = nil
+          @is_set = false
+        end
+      end
+
+      # Invokes the cancellation function provided.
+      def cancel
```
jmuk: This cancel seems not update `@resource`, and that may cause a problem if another thread is on `wait` unconditionally.

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,448 @@
...
+
+      # Checks to see if the event has been set. A set Event signals that
+      # there is data in @result.
+      # @return [Boolean] Whether the event has been set.
+      def set?
+        @is_set
+      end
+
+      # Signifies that the event has been set and that there is data in @result.
+      def set
```
jmuk: This method should be consolidated with `result=`. Otherwise there can be a race condition.

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,448 @@
...
+      end
+
+      # Checks to see if the event has been set. A set Event signals that
+      # there is data in @result.
+      # @return [Boolean] Whether the event has been set.
+      def set?
+        @is_set
+      end
+
+      # Signifies that the event has been set and that there is data in @result.
```
jmuk: You should comment that this method is for internal-use. Event class will be returned to the API users, and something would go wrong if the user invokes this method.

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,448 @@
...
+      def set
+        @mutex.synchronize do
+          @is_set = true
+          @resource.broadcast
+        end
+      end
+
+      # Resets the Event to not being set and clears the @result.
+      def clear
```
jmuk: Not sure what is the purpose of this method. Gax-Python has this because Python's threading.Event has a method like this, please clarify why it needs to be in Ruby too.

lib/google/gax/bundling.rb
jmuk: Also please write the document for wait method

lib/google/gax/bundling.rb
Landrito: I see, adding ```@resource.broadcast``` to the function should solve this problem.

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,448 @@
...
+      'receive all responses'
+    ].join
+
+    # Helper function for #compute_bundle_id.
+    # Used to retrieve a nested field signified by name where dots in name
+    # indicate nested objects.
+    #
+    # @param obj [Object] an object.
+    # @param name [String] a name for a field in the object.
+    # @raise [NoMethodError] if the object does not contain the named attribute.
```
jmuk: nit: now this method does not raise errors (same in compute_bundle_id)

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,448 @@
...
+                     bundled_field,
+                     bundling_request,
+                     subresponse_field: nil)
+        @api_call = api_call
+        @bundle_id = bundle_id
+        @bundled_field = bundled_field
+        @bundling_request = bundling_request
+        @subresponse_field = subresponse_field
+        @in_deque = []
+        @event_deque = []
```
jmuk: They are not deque, better to use other name (event_duque -> events, in_deque -> inputs?)

lib/google/gax/bundling.rb
Landrito: I agree, I assumed it was going to be used elsewhere when I integrated bundling but did not seem to be the case. I will remove this function.

lib/google/gax/bundling.rb
Landrito: I removed set by consolidating it into ```result=```

lib/google/gax/bundling.rb
Landrito: I removed set by consolidating it into ```result=```

lib/google/gax/bundling.rb
Landrito: I went with inputs and events. thanks.

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,445 @@
...
+      # there is data in @result.
+      # @return [Boolean] Whether the event has been set.
+      def set?
+        @is_set
+      end
+
+      # Invokes the cancellation function provided.
+      # The returned cancellation function returns true if all elements
+      # was removed successfully from the inputs, and false if it was not.
+      def cancel
```
Landrito: Might need a second set of eyes on this logic. @jmuk

lib/google/gax/bundling.rb
```diff
@@ -0,0 +1,445 @@
...
+      #
+      # @param api_call [Proc] used to make an api call when the task is run.
+      # @param bundle_id [String] the id of this bundle.
+      # @param bundle_desc [BundleDescriptor] describes the structure of the
+      #     bundled call.
+      # @param bundling_request [Object] the request to pass as the arg to
+      #     the api_call.
+      # @return [Event] an Event that can be used to wait on the response.
+      #
+      # TODO: Implement byte and element count limits.
```
jmuk: nit: it's probably better to move this TODO inside of the method. The comment here will appear as a part of public documentation.

jmuk:
LGTM, with one nit comment.

Please make sure to choose 'Squash and merge' instead of 'merge'.

PR #35, opened 2016-07-27
Use metadata argument

jmuk:
gRPC 0.15.0 actually assumes metadata: argument rather than kwargs.


jmuk:
@geigerj, please take a look when you are not busy.

geigerj:
LGTM

PR #39, opened 2016-08-18
Hide CallSettings as private.

jmuk:
In Ruby, a 'private constant' is a constant which will have
to be accessed without prefix. So all entties inside of
Google::Gax can still access it as-is, but outside the module
it can't be accessed as Google::Gax::CallSettings.

This affects the tests, so this patch introduced a renaming
to toplevel ::CallSettings in a way to avoid violating
access of private constant.

updates https://github.com/googleapis/gax-python/issues/112

### DIFF
--- a/lib/google/gax.rb
+++ b/lib/google/gax.rb
@@ -35,6 +35,7 @@ require 'google/gax/settings'
 require 'google/gax/version'
 
 module Google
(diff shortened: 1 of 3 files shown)


geigerj:
Cool, thanks for doing this!

LGTM

PR #46, opened 2016-10-12
Add longrunning API from gapic generated code

swcloud:


lib/google/longrunning/operations_api.rb
```diff
@@ -0,0 +1,330 @@
...
+# For the short term, the refresh process will only be runnable by Google
+# engineers.
+#
+# The only allowed edits are to method and file documentation. A 3-way
+# merge preserves those additions if the generated source changes.
+
+require "json"
+require "pathname"
+
+require "google/gax"
```
jmuk: Be careful about the circular dependency; when the final design of LRO is implemented, some part of gax library will depend on this file. cc @landrito

lib/google/longrunning/operations_pb.rb
```diff
@@ -0,0 +1,50 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
jmuk: I think we've decided these pb.rb files are either in a part of googleapis-common-protos package or an independent package.

lib/google/gax/version.rb
```diff
@@ -29,6 +29,6 @@
 
 module Google
   module Gax
-    VERSION = '0.5.1'.freeze
+    VERSION = '0.5.2'.freeze
```
jmuk: This type of the change would require an increment on the minor version part, not the patch part (see http://semver.org/).

swcloud: ok. I'll change it to '0.6.0'.

lib/google/longrunning/operations_pb.rb
swcloud: will remove _pb.rb files

jmuk:
Oh, and you'd have to add dependency on `googleapis-common-protos` package.

swcloud:
Added the dependency and fixed two rubocop offenses.

PR #47, opened 2016-10-25
Add LRO wrapper.

landrito:
### Still needed:
- [x] Unit tests (Exponential retrying still needed)
- [x] Finished design review from gcloud-ruby


lib/google/gax/operation.rb
```diff
@@ -0,0 +1,138 @@
...
+    class Operation
+      attr_reader :grpc_op, :client
+
+      # @param grpc_op [Google::Longrunning::Operation]
+      #   The inital longrunning operation.
+      # @param client [Google::Longrunning::OperationsApi]
+      #   The client that handles the grpc operations.
+      def initialize(grpc_op, client: nil)
+        @grpc_op = grpc_op
+        @client = client.nil? Google::Longrunning::OperationsApi.new : client
```
jmuk: seems syntactically invalid? (`client.nil? ? ...`?)

lib/google/gax/operation.rb
```diff
@@ -0,0 +1,138 @@
...
+      end
+
+      # If the operation is done, returns the result, otherwise returns nil.
+      #
+      # @return [nil | Google::Rpc::Status | Object]
+      #   The result of the operation
+      def results
+        return nil unless done?
+        return @grpc_op.error if @grpc_op.result == :error
+        @grpc_op.response.value if @grpc_op.result == :response
```
jmuk: `response` is actually `google.protobuf.Any`. Does protobuf-ruby automatically unpack the result? otherwise, `Google::Protobuf::Any.unpack` might be required.

lib/google/gax/operation.rb
landrito: Fixed

lib/google/gax/operation.rb
landrito: I see. You are correct. I'll update in next revision.

lib/google/gax/operation.rb
```diff
@@ -0,0 +1,163 @@
...
+      #   Since this method changes internal state, it returns itself.
+      def reload!(backoff_settings: nil)
+        options = CallSettings.new(
+          retry_options: RetryOptions.new(
+            backoff_settings: backoff_settings
+          )
+        )
+        @grpc_op = @client.get_operation @grpc_op.name, options: options
+        if done?
+          callbacks.each { |proc| proc.call(results) }
```
blowmage: I would prefer `proc.call(self)` to `proc.call(results)`.

lib/google/gax/operation.rb
```diff
@@ -0,0 +1,163 @@
...
+          delay = [delay * delay_multiplier, max_delay].min
+          reload! backoff_settings
+        end
+        yield(results) if block_given?
+      end
+
+      # Registers a callback to be run when a refreshed operation is marked
+      # as done. If the operation has completed prior to a call to this function
+      # the callback will be called instead of registered.
+      def on_done
```
blowmage: This can be much simpler.

```ruby
def on_done &block
   callbacks.push(block)
end
```

lib/google/gax/operation.rb
landrito: Interesting. 

I am assuming that the user would be doing callbacks as such:
```rb
op = api.method_returns_gax_operation()
op.on_done do |result|
  # Handle result
end
```

I think I see your point though. By using self it allows the user to handle errors.
```rb
op = api.method_returns_gax_operation()
op.on_done do |operation|
  raise 'Error! Oh no!' if operation.error?
  results = operation.results
  # Handle results.
end
```

Correct me if theres errors in my understanding.

blowmage: Yep, that's my concern. There is also meta data that we can't infer from the results, such as operation id/name, that might be useful as well.

lib/google/gax/operation.rb
```diff
@@ -0,0 +1,167 @@
...
+          reload!(backoff_settings: backoff_settings)
+        end
+        yield(results) if block_given?
+      end
+
+      # Registers a callback to be run when a refreshed operation is marked
+      # as done. If the operation has completed prior to a call to this function
+      # the callback will be called instead of registered.
+      def on_done(&block)
+        if done?
```
jmuk: I think this could address a theoretical situation where the initial return value of Operation could have `done: true` already.  I don't know this happens actually (probably quite unlikely) but if that happens, the `on_done` block won't run at all, which makes me a bit nervous.

spec/google/gax/operation_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+      mock_client = MockLroClient.new(get_method: get_method)
+      op = create_op(GrpcOp.new(done: false), client: mock_client)
+      begin
+        op.wait_until_done!(backoff_settings: backoff_settings)
+      rescue Google::Gax::RetryError => exc
+        expect(exc).to be_a(Google::Gax::RetryError)
+      end
+    end
+
+    # TODO: Test exponential backoff.
```
bjwatson: Include a link to an issue number for fixing this, unless you plan to do it in this PR

lib/google/gax/operation.rb
```diff
@@ -0,0 +1,167 @@
...
+      #
+      # @param [Class] The class type to be unpacked from the response.
+      # @return [nil | Google::Rpc::Status | Object | Google::Protobuf::Any ]
+      #   The result of the operation
+      def results(response_type: nil)
+        return nil unless done?
+        return @grpc_op.error if error?
+        return @grpc_op.response.unpack(response_type) if response_type
+        @grpc_op.response
+      end
```
landrito: After thinking about how responses can be unpacked in dynamic languages [here](https://github.com/googleapis/gax-nodejs/pull/64#discussion_r86889911). It may be good to make this response unpacked based on the type returned from the response. I think it can be implemented as follows:

```rb
def results (response_type: nil)
  return nil unless done?
  return @grpc_op.error if error?
  return @grpc_op.response.unpack(response_type) if response_type
  begin
    ResponseType = Google::Protobuf::DescriptorPool.generated_pool.lookup(@grpc_op.type_name).msgclass
    return @grpc_op.response.unpack(ResponseType)
  rescue
    puts "Operation response type not loaded. Unable to unpack."
  end
  @grpc_op.response
end
```

@jmuk thoughts?

landrito:
Added some logic to unpack results based on the response. And the test for exponential backoff. @jmuk

spec/google/gax/operation_spec.rb
landrito: I fixed it in this PR!

lib/google/gax/operation.rb
```diff
@@ -0,0 +1,224 @@
...
+        rescue RuntimeError => e
+          warn e.message + ' The raw response was returned. To get the \
+               unpacked response object, either specify the type, \
+               or require the protofile containing the type: ' +
+               @grpc_op.response.type_name + '.'
+        end
+        @grpc_op.response
+      end
+
+      def metadata(type: nil)
```
swcloud: Documentation for this method?

lib/google/gax/operation.rb
```diff
@@ -0,0 +1,224 @@
...
+    #   @return [Google::Longrunning::Operation] The wrapped grpc
+    #     operation object.
+    # @attribute [r] client
+    #   @return [Google::Longrunning::OperationsApi] The client that handles the
+    #     grpc operations.
+    # @attribute [rw] call_options
+    #   @return [Google::Gax::CallOptions] The call options used when reloading
+    #     the operation.
+    class Operation
+      attr_reader :grpc_op, :client
```
swcloud: Is it necessary to expose ':client' here?

landrito: Nope. Fixed.

lib/google/gax/operation.rb
landrito: Added.

lib/google/gax/operation.rb
```diff
@@ -0,0 +1,224 @@
...
+      # If the operation is not finished then this will return false.
+      #
+      # @return [Boolean] Whether an error has been returned.
+      def error?
+        done? ? @grpc_op.result == :error : false
+      end
+
+      # Cancels the operation.
+      def cancel
+        @client.cancel_operation @grpc_op.name
```
landrito: Fixed! Thanks for the review!

landrito: Fixed. Thanks for the review!

lib/google/gax/operation.rb
```diff
@@ -0,0 +1,240 @@
...
+    #   op.done?
+    #   op.reload!
+    #   op.done?
+    #   results = op.results
+    #   raise results if op.error?
+    #
+    # @attribute [r] grpc_op
+    #   @return [Google::Longrunning::Operation] The wrapped grpc
+    #     operation object.
+    # @attribute [r] client
```
swcloud: client is not an attribute anymore.

lib/google/gax/operation.rb
```diff
@@ -0,0 +1,240 @@
...
+require 'google/protobuf/well_known_types'
+
+module Google
+  module Gax
+    # A class used to wrap Google::Longrunning::Operation objects.
+    #
+    # @example General usage example
+    #   require 'google/gax/operation'
+    #
+    #   op = Google::Gax::Operation.new api.methodThatReturnsOperation(name)
```
swcloud: The sample constructor use does not seem to be consistent with the real definition.

lib/google/gax/operation.rb
```diff
@@ -0,0 +1,240 @@
...
+      # If the operation is done, returns the result, otherwise returns nil.
+      # If the operation response is an error, the error will be returned.
+      # If a type is provided, the response will be unpacked using the type
+      # provided; returning nil if the response is not of the type provided.
+      # If the type is not of provided, the response will be unpacked using
+      # the response's type_url if the type_url is found in the
+      # Google::Protobuf::DescriptorPool.generated_pool.
+      # If the type cannot be found the raw response is retuned.
+      #
+      # @param type [Class]] The class type to be unpacked from the response.
```
swcloud: Extra close bracket.

lib/google/gax/operation.rb
```diff
@@ -0,0 +1,240 @@
...
+
+      # Returns the metadata of an operation. If a type is provided,
+      # the metadata will be unpacked using the type provided; returning nil
+      # if the metadata is not of the type provided.
+      # If the type is not of provided, the metadata will be unpacked using
+      # the metadata's type_url if the type_url is found in the
+      # Google::Protobuf::DescriptorPool.generated_pool.
+      # If the type cannot be found the raw metadata is retuned.
+      #
+      # @param type [Class]] The class type to be unpacked from the response.
```
swcloud: Extra close bracket

lib/google/gax/operation.rb
swcloud: I suggest writing more complete sample usage cases for this class including on_done, wait_until_done?

landrito: Fixed!

lib/google/gax/operation.rb
landrito: Fixed!

lib/google/gax/operation.rb
landrito: Fixed!

PR #49, opened 2016-11-22
Updates to LROs unpacking strategy.

landrito:
### Unpacking strategy.
Previously I thought it would be a clever idea to unpack the operation result and metadata based on the `type_url` found in the `google.protobuf.Any` message. Toolkit has been updated to have the api producer explicitly specify the expected types of the result and metadata. Since these types are explicitly specified by the api producer, these should be the only types unpacked from the operations.

This resulted in having the result and metadata types specified in the constructor of the `Google::Gax::Operation` rather than being dynamically found in the `type_url` of the `google.protobuf.Any` message or being specified by the user when the results or metadata messages are called.

### DIFF
--- a/lib/google/gax/operation.rb
+++ b/lib/google/gax/operation.rb
@@ -148,7 +148,7 @@ module Google
       # If the type cannot be found the raw metadata is retuned.
       #
       # @return [Object, nil]
-      #   The result of the operation. Can be nil.
+      #   The metadata of the operation. Can be nil.
       def metadata
         return nil if @grpc_op.metadata.nil?
         @grpc_op.metadata.unpack(@metadata_type)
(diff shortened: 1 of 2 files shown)


lib/google/gax/operation.rb
```diff
@@ -152,22 +147,11 @@ def results(type: nil)
       # Google::Protobuf::DescriptorPool.generated_pool.
       # If the type cannot be found the raw metadata is retuned.
       #
-      # @param type [Class] The class type to be unpacked from the response.
-      #
-      # @return [nil | Object | Google::Protobuf::Any ]
-      #   The result of the operation
-      def metadata(type: nil)
+      # @return [Object, nil]
+      #   The result of the operation. Can be nil.
```
landrito: Done.

landrito:
Thanks for the review @swcloud!

PR #53, opened 2017-01-05
Ported unpack from well known types to resolve protobuf dependency error.

landrito:

PR #55, opened 2017-01-27
Revert "Ported unpack from well known types to resolve protobuf dependency error."

landrito:
Reverts googleapis/gax-ruby#53

Looks like a rubocop error. Looking into it.


landrito:
This is strange. Revert is also failing.

PR #60, opened 2017-02-16
Fix retry_codes bug in config json

swcloud:


lib/google/gax/settings.rb
```diff
@@ -400,9 +400,11 @@ def upper_camel_to_lower_underscore(string)
     #   {
     #     "interfaces": {
     #       "google.fake.v1.ServiceName": {
-    #         "retry_codes": {
-    #           "idempotent": ["UNAVAILABLE", "DEADLINE_EXCEEDED"],
-    #           "non_idempotent": []
+    #         "retry_codes_def": {
```
landrito: `retry_codes_def` is the inner key not the outer key. [Example](https://github.com/googleapis/gax-ruby/blob/master/lib/google/longrunning/operations_client_config.json#L5)

blowmage: It seems odd to me to have `retry_codes_def` nested under `retry_codes`. There isn't a `_def` node nested under `retry_params`, and there isn't a `_def` node nested under `methods`. Why is `retry_codes` different?

swcloud: @blowmage We're still tracking down why 'retry_codes_def' was used.

blowmage: Awesome, thanks.

swcloud: We decided to remove 'retry_codes_def' from all json files. So I'm closing this PR.

blowmage:
Woo! 🎉

PR #62, opened 2017-02-24
LRO client updates

landrito:


lib/google/gax/version.rb
```diff
@@ -29,6 +29,6 @@
 
 module Google
   module Gax
-    VERSION = '0.8.0'.freeze
+    VERSION = '0.9.0'.freeze
```
blowmage: Can this be 0.8.1 instead of 0.9.0? It looks to be backwards compatible and I'd like to not have to release new versions of the google-cloud-* gems again

landrito: Ah It escaped my mind that I purposefully made the http headers change backwards compatible. Will do.

landrito: Done

swcloud: I think we might take this opportunity to  force the client library to upgrade to have the http header tracking capability when they upgrade google-gax? for example, all the gemspec should rely on 0.8.1 or 0.9.0?

lib/google/longrunning/operations_client.rb
```diff
@@ -96,17 +90,27 @@ def initialize \
...
-          "#{CODE_GEN_NAME_VERSION} gax/#{Google::Gax::VERSION} " \
-          "ruby/#{RUBY_VERSION}".freeze
+
+        if app_name || app_version
+          warn "`app_name` and `app_version` are no longer being used in the request headers."
+        end
+
+        google_api_client = "gl-ruby/#{RUBY_VERSION}"
+        google_api_client << " #{lib_name}/#{lib_version}" if lib_name
+        google_api_client << " gapic/0.1.0 gax/#{Google::Gax::VERSION}"
```
bjwatson: Please don't report a GAPIC version unless it's real. That will lead to bad data. If we cannot currently determine the GAPIC version here without a lot of refactoring, then just do `gapic/` (without the version).

bjwatson: Arguably in this case, there is no GAPIC so maybe we can just skip the `gapic/` token.

blowmage: I assumed the GAPIC version here was the version of the code that generated this file. Is that not the case?

bjwatson: Oh, right. Is this generated code @landrito? In that case, I have no concerns about this.

bjwatson: Something that makes me think this is fake is that the version is `0.1.0`. @landrito, @swcloud  are we bumping this when we regenerate the GAPIC code? If not, we won't be able to reasonably track what codegen versions are being used.

PR #75, opened 2017-06-06
Add protobuf type coercion of native Hash

landrito:


landrito:
Hm looks like CI Rubocop is failing on some lines I didn't touch. I remember @jmuk had an idea why this happens.

spec/google/gax/util_pb.rb
```diff
@@ -0,0 +1,54 @@
+# Copyright 2017, Google Inc.
```
geigerj: Is this generated by `protoc` or handwritten? If it's generated, let's check in the proto file as well ([like Python](https://github.com/googleapis/gax-python/tree/master/tests/fixtures)), which makes it easier to modify in the future. Otherwise, let's replace this with a generated file.

landrito: Done!

jmuk:
yeah, the current gemspec allows to install rubocop with minor version update, and it sometimes starts reporting new things which weren't reported.

I think:
- for this PR, it's okay to leave it as-is.
- for the long-term, probably we might want to stick the rubocop version -- currenntly it's specified as `'~> 0.32'` in `google-gax.gemspec`, which means `>= 0.32` and `< 1.0`. Could be `~> 0.49.0` (i.e. `>= 0.49.0` and `< 0.50.0`) or `0.49.1` (stick with a certain version)

landrito:
Fixed the rubocop version to the version of the most recent pr pass `0.47.1` specifically. Looks like it worked. Thanks @jmuk!! 🎊

PR #76, opened 2017-06-21
Port credentials class from google-cloud-core

landrito:
Updates: https://github.com/googleapis/toolkit/issues/1324

#### Deviations from google-cloud-core
- added `:updater_proc` as a delegated method
- removed default gcloud paths since they don't make sense in the google-gax context
- changed the test file to use rspec instead of minitest

### DIFF
--- a/lib/google/gax/credentials.rb
+++ b/lib/google/gax/credentials.rb
@@ -111,9 +111,8 @@ module Google
 
       # Verify that the keyfile argument is provided.
       def verify_keyfile_provided!(keyfile)
-        if keyfile.nil?
-          raise 'The keyfile passed to Google::Gax::Credentials.new was nil.'
-        end
+        return unless keyfile.nil?
+        raise 'The keyfile passed to Google::Gax::Credentials.new was nil.'
       end
 
       # Verify that the keyfile argument is a file.
--- a/lib/google/gax/credentials.rb
+++ b/lib/google/gax/credentials.rb
@@ -1,0 +1,149 @@
+# Copyright 2017, Google Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met:
+#
+#     * Redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above
+# copyright notice, this list of conditions and the following disclaimer
+# in the documentation and/or other materials provided with the
+# distribution.
+#     * Neither the name of Google Inc. nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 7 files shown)


lib/google/gax/credentials.rb
```diff
@@ -0,0 +1,149 @@
...
+          json['scope'] ||= scope
+          @client = init_client json
+        end
+        @client.fetch_access_token!
+      end
+
+      ##
+      # Returns the default credentials.
+      #
+      def self.default(scope: nil)
```
geigerj: Possible to add a test that subclasses this like the autogen code will to get test coverage here?

lib/google/gax/credentials.rb
```diff
@@ -0,0 +1,149 @@
...
+        scope ||= self::SCOPE
+        client = Google::Auth.get_application_default scope
+        new client
+      end
+
+      protected
+
+      # Verify that the keyfile argument is provided.
+      def verify_keyfile_provided!(keyfile)
+        raise 'You must provide a keyfile to connect with.' if keyfile.nil?
```
geigerj: When could this get raised? It seems wrong to tell users they "must" provide a keyfile given that for GAPIC, we expect most people to use ADC, and the keyfile to be a more advanced use case.

lib/google/gax/credentials.rb
landrito: Yep! Done.

lib/google/gax/credentials.rb
landrito: I don't see this getting raised for our cases. I made the error message less aggressive.

landrito:
Heres the Toolkit PR: https://github.com/googleapis/toolkit/pull/1387

landrito:
@geigerj this is ready for another round of review.

lib/google/gax/grpc.rb
```diff
@@ -72,22 +74,23 @@ module Grpc
...
                       channel: nil,
+                      chan_creds: nil,
                       updater_proc: nil,
                       scopes: nil)
         address = "#{service_path}:#{port}"
-        if channel.nil?
-          chan_creds = GRPC::Core::ChannelCredentials.new if chan_creds.nil?
+        if channel
+          yield(address, nil, channel_override: channel)
+        elsif chan_creds
```
geigerj: I think the old way was right, and this is wrong. It's totally OK to specify both a `chan_creds` (credentials on the channel, like SSL) and an updater proc (credentials on the call, like OAuth).

landrito: I see. The reason for this change is because in toolkit if we consolidate the updater_proc and the channel_creds into a single parameter. For the user to be able to customize both a chan_creds and updater proc, they would have to compose their chan_creds themselves. If we leave the behavior as such. Their composed chan_creds would get overwritten by the default updater_proc.

geigerj: I see. SGTM as long as we document on the GAPIC how to compose the chan_creds and pass it to the client.

landrito: SGTM.

geigerj: Maybe raise an exception if multiple overriding params are specified instead of silently ignoring some, though.

landrito: Done.

landrito:
Raised an exception for that case and added some unit tests for Google::Gax::Grpc#create_stub.

Issue #54, opened 2017-01-20
File permissions cause LoadError


blowmage:
Hi, the google-cloud-ruby project has a couple of reports of `LoadError` when requiring "google/gax". (See GoogleCloudPlatform/google-cloud-ruby#1193 and GoogleCloudPlatform/google-cloud-ruby#1194) The cause for this is that the gems are packaged with files using 640 permissions, where 664 or 644 permissions are expected.

A very common production scenario for Ruby applications is to have a `deploy` user install the application and all dependent libraries, and an `app` user account run the application. Because of how the google-gax file permissions are set, it is entirely possible that the `app` user cannot read the files, and a `LoadError` is raised.

Here are the permissions for google-gax 0.6.0:

```
├── [-rw-r-----]  README.md
├── [-rw-r-----]  Rakefile
├── [drwxr-xr-x]  lib
│   └── [drwxr-xr-x]  google
│       ├── [drwxr-xr-x]  gax
│       │   ├── [-rw-r-----]  api_callable.rb
│       │   ├── [-rw-r-----]  bundling.rb
│       │   ├── [-rw-r-----]  errors.rb
│       │   ├── [-rw-r-----]  grpc.rb
│       │   ├── [-rw-r-----]  path_template.rb
│       │   ├── [-rw-r-----]  settings.rb
│       │   └── [-rw-r-----]  version.rb
│       ├── [-rw-r-----]  gax.rb
│       └── [drwxr-xr-x]  longrunning
│           ├── [-rw-r-----]  operations_api.rb
│           └── [-rw-r-----]  operations_client_config.json
└── [drwxr-xr-x]  spec
    ├── [drwxr-xr-x]  google
    │   └── [drwxr-xr-x]  gax
    │       ├── [-rw-r-----]  api_callable_spec.rb
    │       ├── [-rw-r-----]  bundling_spec.rb
    │       ├── [-rw-r-----]  path_template_spec.rb
    │       └── [-rw-r-----]  settings_spec.rb
    └── [-rw-r-----]  spec_helper.rb

7 directories, 17 files
```

This problem also affects googleapis-common-protos, but I don't see a repo for that gem, so I'm mentioning it here.

blowmage:
Likely related to #42.

blowmage:
This problem does not affect any of the google-cloud-* gems, googleauth, google-protobuf, grpc, or grpc-google-iam-v1.

blowmage:
[...]

PR #92, opened 2017-09-27
Fix file permissions

geigerj:
Grant a+rx permissions throughout to prevent re-occurrence of #54


geigerj:
Right... and git doesn't seem to track directories permissions :/

landrito:
Maybe it might be better to write a publishing script that makes sure that the directories and files have correct permissions before publishing?

PR #93, opened 2017-09-28
GAX bug fixes

geigerj:
- Disables error detail deserialization until a fix in gRPC is released.
- Allows instances of proto messages to be passed to to_proto
- Bump version. Note that non-sequential increase is due to rollbacks

### DIFF
--- a/lib/google/gax/util.rb
+++ b/lib/google/gax/util.rb
@@ -42,12 +42,15 @@ module Google
     # messages using hashes but does not allow for nested hashes to instantiate
     # nested submessages.
     #
-    # @param hash [Hash] The hash to be converted into a proto message.
+    # @param hash [Hash || Class] The hash to be converted into a proto message.
+    #   If an instance of the proto message class is given, it is returned
+    #   unchanged.
     # @param message_class [Class] The corresponding protobuf message class of
     #   the given hash.
     #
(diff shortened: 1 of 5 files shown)


lib/google/gax/util.rb
```diff
@@ -42,12 +42,15 @@ module Gax
...
-    # @param hash [Hash] The hash to be converted into a proto message.
+    # @param hash [Hash || Class] The hash to be converted into a proto message.
+    #   If an instance of the proto message class is given, it is returned
+    #   unchanged.
     # @param message_class [Class] The corresponding protobuf message class of
     #   the given hash.
     #
     # @return [Object] An instance of the given message class.
     def to_proto(hash, message_class)
+      return hash if hash.is_a? message_class
```
landrito: Would be good to add a check that the hash is a hash after this line.

geigerj: Done.

lib/google/gax/version.rb
```diff
@@ -29,6 +29,6 @@
 
 module Google
   module Gax
-    VERSION = '0.8.8'.freeze
+    VERSION = '0.8.11'.freeze
```
landrito: I assume the version changing by 3 is because of bad file permissions?

PR #97, opened 2017-10-26
Support googleauth 0.6.1

geigerj:


landrito:
Should this PR also remove the Google::Gax::Credentials class as well?

geigerj:
@landrito Indeed, thanks for the catch! That actually means we need to regenerate the LRO client, as well. I'll update when this is ready for review.

geigerj:
Credentials removed.

geigerj:
Hmm... although the new Operations client doesn't seem to `require` googleauth, which seems wrong. (It also doesn't currently have test coverage, although I'd rather address that on follow-up than on this PR.) Taking a look.

geigerj:
It looks like this is OK. (Requiring `google/gax` pulls in `googleauth` transitively through the `grpc` module.) CI failure seems to be due to gRPC 1.7.1 not supporting Ruby 2.0.0. @landrito PTAL?

landrito:
I think it's better to not have a transitive dependency to googleauth: https://github.com/googleapis/toolkit/pull/1665

lib/google/longrunning/operations_client.rb
```diff
@@ -139,14 +131,14 @@ def initialize \
...
           credentials ||= updater_proc
         end
-        if app_name || app_version
-          warn "`app_name` and `app_version` are no longer being used in the request headers."
+        if service_path != SERVICE_ADDRESS || port != DEFAULT_SERVICE_PORT
+          warn "`service_path` and `port` parameters are deprecated and will be removed"
         end
 
-        credentials ||= Google::Gax::Credentials.default(scope: scopes)
+        credentials ||= Google::Auth::Credentials.default
```
landrito: This is wrong only for LRO generation since LRO _needs_ to have scopes passed into it or else the channel will not have a valid endpoint to reach. I'll update https://github.com/googleapis/toolkit/pull/1665 to fix this.

geigerj: Updated.

PR #104, opened 2017-11-13
GaxError presents BadStatus values

blowmage:
This PR relays the underlying BadStatus values `code`, `details`, and `metadata` without resorting to dynamic hackery using `method_missing`. If the underlying error object does not have values, or is not a BadStatus, `nil` is returned.

[refs GoogleCloudPlatform/google-cloud-ruby#1791]
[refs GoogleCloudPlatform/google-cloud-ruby#1825]

### DIFF
--- a/lib/google/gax/errors.rb
+++ b/lib/google/gax/errors.rb
@@ -35,7 +35,7 @@ module Google
   module Gax
     # Common base class for exceptions raised by GAX.
     class GaxError < StandardError
-      attr_reader :details
+      attr_reader :status_details
 
       # @param msg [String] describes the error that occurred.
       def initialize(msg)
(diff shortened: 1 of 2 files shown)


lib/google/gax/errors.rb
```diff
@@ -53,6 +54,21 @@ def initialize(msg)
           @cause
         end
       end
+
+      def code
+        return nil unless cause && cause.respond_to?(:metadata)
```
geigerj: `cause.respond_to?(:metadata)` -> `cause.respond_to?(:code)`?

Similarly with `details`.

blowmage: Oops. Fixed.

blowmage:
Build is fixed. Thank you for your patience.

PR #109, opened 2017-11-20
Allow operations client to be subclassed

geigerj:
Note: this is hand-edited for now to preserve compatibility with older
GAPIC clients that set the service address through the ctor. We will
refresh the LRO codegen entirely for the 1.0 release.


geigerj:
@landrito Could you take a look at this one, too? It's required for https://github.com/googleapis/toolkit/pull/1702.

Issue #86, opened 2017-08-22
CallOptions kwargs doesn't work


geigerj:
### What

The CallOptions kwarg seems to be passed to `metadata` instead of being used as arbitrary optional args. See: https://github.com/googleapis/gax-ruby/blob/e8c66e40fe2881cf9b606f33acf4e8b7d13cefdc/lib/google/gax/api_callable.rb#L376:

### Why

Users may want to use gRPC optional arguments that we don't explicitly support through GAX. In particular, I encountered this bug trying to use the gRPC setting `return_op`:

```
require 'google/cloud/logging'

client = Google::Cloud::Logging::V2::LoggingServiceV2Client.new
options = Google::Gax::CallOptions.new kwargs: {return_op: true}
client.list_logs 'projects/gapic-test', options: options

# Results in:
#   Google::Gax::RetryError: GaxError Exception occurred in retry method that was not classified as transient, caused by Header values must be of type string or array
#   ...
```

landrito:
Ack. Can you add a priority label to this?

geigerj:
Unfortunately not, but treat it as P2. (I'll get Luke to give me admin permissions so that I can create priority labels.)

jbolinger:
I'm not sure that there is currently a way to pass "return_op: true" to the grpc layer if we want to allow this. kwargs are passed as the metadata parameter from gax and I don't see a way to set return_op:

https://github.com/grpc/grpc/blob/0819ff56f5b7bed6a351db8913d7bb75be0e4d95/src/ruby/lib/grpc/generic/service.rb#L170

geigerj:
@jbolinger Right. The issue is that we should *not* be passing kwargs as the metadata parameter; this is a mistake. Rather, it should be "double splatted" out into arbitrary keyword arguments of the gRPC stub method.

That is,

```
my_gapic_client.my_rpc(request, Google::Gax::CallOptions.new(
  kwargs: {
    metadata: { 'x-some-header' => 'some value' },
    return_op: true,
    some_other_parameter: some_other_argument
  }
))
```

should trigger the gRPC stub call

```
my_grpc_stub.my_rpc(
  ...,
  metadata: { 'x-some-header' => 'some value' },
  return_op: true,
  some_other_parameter: some_other_argument
)
```

geigerj:
Note: a decent amount of code in google-cloud-ruby relies on this buggy behavior, so we should be very careful with the release.

[...]

PR #118, opened 2018-02-22
Update metadata handling and add block to expose grpc ops internally

jbolinger:
This is a WIP for a potential new toolkit feature. The main goal is to invoke all internal grpc calls using the return_op=true flag so that any additional grpc information is available and can be trapped in gax, via an optional block. Those blocks will then be used by the generated gapic methods to trap and expose things like response metadata to the caller.

Related toolkit change: https://github.com/googleapis/toolkit/pull/1865

Note: Do not merge until the feature is finalized!

### DIFF
--- a/lib/google/gax/api_callable.rb
+++ b/lib/google/gax/api_callable.rb
@@ -136,15 +136,15 @@ module Google
       # @param request [Object]
       #   The initial request object.
       # @param settings [CallSettings]
-      #   The call settings to enumerate pages.
+      #   The call settings to enumerate pages
       # @return [PagedEnumerable]
       #   returning self for further uses.
-      def start(api_call, request, settings)
+      def start(api_call, request, settings, block)
         @func = api_call
         @request = request
         page_token = settings.page_token
         @request[@request_page_token_field] = page_token if page_token
-        @page = @page.dup_with(@func.call(@request))
+        @page = @page.dup_with(@func.call(@request, block))
         self
       end
 
@@ -222,8 +222,8 @@ module Google
     # @raise [StandardError] if +settings+ has incompatible values,
     #   e.g, if bundling and page_streaming are both configured
     def create_api_call(func, settings, params_extractor: nil)
-      api_caller = proc do |api_call, request|
-        api_call.call(request)
+      api_caller = proc do |api_call, request, settings, block|
+        api_call.call(request, block)
       end
 
       if settings.page_descriptor
@@ -239,7 +239,7 @@ module Google
         api_caller = bundleable(settings.bundle_descriptor)
       end
 
-      proc do |request, options = nil|
+      proc do |request, options = nil, &block|
         this_settings = settings.merge(options)
         if params_extractor
           params = params_extractor.call(request)
@@ -247,13 +247,13 @@ module Google
         end
         api_call = if settings.retry_codes?
                      retryable(func, this_settings.retry_options,
-                               this_settings.kwargs)
+                               this_settings.metadata)
                    else
                      add_timeout_arg(func, this_settings.timeout,
-                                     this_settings.kwargs)
+                                     this_settings.metadata)
                    end
         begin
-          api_caller.call(api_call, request, this_settings)
+          api_caller.call(api_call, request, this_settings, block)
         rescue *settings.errors => e
           error_class = Google::Gax.from_error e
           raise error_class.new('RPC failed')
@@ -277,12 +277,13 @@ module Google
     # @return [Proc] A proc takes the API call's request and returns
     #   an Event object.
     def bundleable(desc)
-      proc do |api_call, request, settings|
-        return api_call(request) unless settings.bundler
+      proc do |api_call, request, settings, block|
+        return api_call(request, block) unless settings.bundler
         the_id = Google::Gax.compute_bundle_id(
           request,
           desc.request_discriminator_fields
         )
+        # TODO: how to handle the block arg in next line?
         settings.bundler.schedule(api_call, the_id, desc, request)
       end
     end
@@ -316,7 +317,7 @@ module Google
     def with_routing_header(settings, params)
       routing_header = params.map { |k, v| "#{k}=#{v}" }.join('&')
       options = CallOptions.new(
-        kwargs: { 'x-goog-request-params' => routing_header }
+          metadata: { 'x-goog-request-params' => routing_header }
       )
       settings.merge(options)
     end
@@ -328,8 +329,9 @@ module Google
     # @param retry_options [RetryOptions] Configures the exceptions
     #   upon which the proc should retry, and the parameters to the
     #   exponential backoff retry algorithm.
+    # @param metadata [Hash] request metadata headers
     # @return [Proc] A proc that will retry on exception.
-    def retryable(a_func, retry_options, kwargs)
+    def retryable(a_func, retry_options, metadata)
       delay_mult = retry_options.backoff_settings.retry_delay_multiplier
       max_delay = (retry_options.backoff_settings.max_retry_delay_millis /
                    MILLIS_PER_SECOND)
@@ -339,13 +341,19 @@ module Google
       total_timeout = (retry_options.backoff_settings.total_timeout_millis /
                        MILLIS_PER_SECOND)
 
-      proc do |request|
+      proc do |request, block|
         delay = retry_options.backoff_settings.initial_retry_delay_millis
         timeout = (retry_options.backoff_settings.initial_rpc_timeout_millis /
                    MILLIS_PER_SECOND)
         deadline = Time.now + total_timeout
         begin
-          a_func.call(request, deadline: Time.now + timeout, metadata: kwargs)
+          op = a_func.call(request,
+                           deadline: Time.now + timeout,
+                           metadata: metadata,
+                           return_op: true)
+          res = op.execute
+          block.call op if block
+          res
         rescue => exception
           unless exception.respond_to?(:code) &&
                  retry_options.retry_codes.include?(exception.code)
@@ -372,10 +380,17 @@ module Google
     # @param a_func [Proc] a proc to be updated
     # @param timeout [Numeric] to be added to the original proc as it
     #   final positional arg.
+    # @param metadata [Hash] request metadata headers
     # @return [Proc] the original proc updated to the timeout arg
-    def add_timeout_arg(a_func, timeout, kwargs)
-      proc do |request|
-        a_func.call(request, deadline: Time.now + timeout, metadata: kwargs)
+    def add_timeout_arg(a_func, timeout, metadata)
+      proc do |request, block|
+        op = a_func.call(request,
+                         deadline: Time.now + timeout,
+                         metadata: metadata,
+                         return_top: true)
+        res = op.execute
+        block.call op if block
+        res
       end
     end
 
--- a/lib/google/gax/api_callable.rb
+++ b/lib/google/gax/api_callable.rb
@@ -222,7 +222,7 @@ module Google
     # @raise [StandardError] if +settings+ has incompatible values,
     #   e.g, if bundling and page_streaming are both configured
     def create_api_call(func, settings, params_extractor: nil)
-      api_caller = proc do |api_call, request, settings, block|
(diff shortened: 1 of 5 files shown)


landrito:
Interesting! Can you provide an example of how a user would use this feature?

jbolinger:
Yes, sorry I meant to do that but forgot!

So, the end user would do something like this:

```ruby
output = client.batch_annotate_images(
    [{
         image: {
             source: { image_uri: "gs://gapic-toolkit/President_Barack_Obama.jpg" }
         },
         features: [{ type: :FACE_DETECTION }]
     }],
    options: Google::Gax::CallOptions.new(metadata: { 'x-some-header' => 'some value' })
  )

puts "output is: #{output}"
puts "response metadata are: #{client.last_response_headers}"
```

The GAPIC layer will be modified to do something like this:
```ruby
def batch_annotate_images requests, options: nil
    req = {
      requests: requests
    }.delete_if { |_, v| v.nil? }
    req = Google::Gax::to_proto(req, Google::Cloud::Vision::V1::BatchAnnotateImagesRequest)
    @batch_annotate_images.call(req, options) { |op| @last_response_headers = op.metadata }
end
```

jbolinger:
I'd also like to enable the following, or something similar, but have to think it through a little more so we don't break the old way and it's consistent with the other APIs

```ruby
op = client.batch_annotate_images(...) do | response, metadata |
   # do stuff with the result...
done
op.wait_until_done!
```

lib/google/gax/api_callable.rb
```diff
@@ -372,10 +380,17 @@ def retryable(a_func, retry_options, kwargs)
...
     # @return [Proc] the original proc updated to the timeout arg
-    def add_timeout_arg(a_func, timeout, kwargs)
-      proc do |request|
-        a_func.call(request, deadline: Time.now + timeout, metadata: kwargs)
+    def add_timeout_arg(a_func, timeout, metadata)
+      proc do |request, block|
+        op = a_func.call(request,
+                         deadline: Time.now + timeout,
+                         metadata: metadata,
+                         return_top: true)
```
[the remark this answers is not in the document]
jbolinger: No idea. I'd be surprised if it made a difference. I can benchmark it using the sample I created Do we have any benchmarks already written?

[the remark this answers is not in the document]
geigerj: @pongad has worked on Spanner benchmarks (not available in open source), which we may be able to re-use. Otherwise, just spamming a high-throughput API should be good enough. (`batch_annotate_images` is probably not a good choice because of the high expected server latency. The Logging API is probably a good test target.)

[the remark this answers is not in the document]
jbolinger: I looked into this a bit more and the library code that's invoked is the same in both cases, so there should not be any difference.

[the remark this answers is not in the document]
geigerj: The original comment (spelling) still needs to be addressed.

lib/google/gax/api_callable.rb
```diff
@@ -277,12 +277,13 @@ def create_api_call(func, settings, params_extractor: nil)
...
     def bundleable(desc)
-      proc do |api_call, request, settings|
-        return api_call(request) unless settings.bundler
+      proc do |api_call, request, settings, block|
+        return api_call(request, block) unless settings.bundler
         the_id = Google::Gax.compute_bundle_id(
           request,
           desc.request_discriminator_fields
         )
+        # TODO: how to handle the block arg in next line?
```
geigerj: Bundling calls should probably raise a runtime exception if you try to get metadata through a block. There's not a 1-1 correspondence between bundled calls and gRPC calls, so getting this right is probably a little tricky.

lib/google/gax/api_callable.rb
```diff
@@ -222,8 +222,8 @@ def page_token
     # @raise [StandardError] if +settings+ has incompatible values,
     #   e.g, if bundling and page_streaming are both configured
     def create_api_call(func, settings, params_extractor: nil)
-      api_caller = proc do |api_call, request|
-        api_call.call(request)
+      api_caller = proc do |api_call, request, _settings, block|
+        api_call.call(request, block)
       end
 
       if settings.page_descriptor
```
geigerj: Passing a block for metadata should probably raise a runtime exception for paginated calls.

spec/google/gax/api_callable_spec.rb
```diff
@@ -71,10 +76,29 @@ module Gax
...
+      allow(op).to receive(:execute) { 2 + op.request + adder }
+
+      func = proc do |request, _deadline: nil, **_kwargs|
+        expect(request).to eq(3)
+        op
+      end
+
+      my_callable = Google::Gax.create_api_call(func, settings)
+      expect(my_callable.call(3)).to eq(5)
+      expect(my_callable.call(3) { adder = 5 }).to eq(5)
```
geigerj: This test is confusing, since you have two different `adder` variables in two different scopes, and the `adder` in the block in this line just gets discarded.

spec/google/gax/api_callable_spec.rb
```diff
@@ -93,15 +117,27 @@ module Gax
     end
 
     it 'iterates over elements' do
-      my_callable = Google::Gax.create_api_call(func, settings)
+      func2 = proc do |request, **kwargs|
```
geigerj: Can you just change `func` to return a mocked `op` instead of wrapping it in `func2` (here and below)?

geigerj:
I still think that passing `return_op: true` for every gRPC call, including the majority of calls where we don't care about the trailers and just retrieve the operation result, seems ugly. But per offline discussion, I won't block the review over it.

Just for confirmation, @apolcyn -- is there any downside to our libraries setting `return_op: true` and blocking on the result for calls where we don't actually need the operation, as opposed to not setting `return_op`?

lib/google/gax/api_callable.rb
```diff
@@ -239,21 +239,21 @@ def create_api_call(func, settings, params_extractor: nil)
...
+      proc do |request, options = nil, &block|
         this_settings = settings.merge(options)
         if params_extractor
           params = params_extractor.call(request)
           this_settings = with_routing_header(this_settings, params)
         end
         api_call = if settings.retry_codes?
                      retryable(func, this_settings.retry_options,
-                               this_settings.kwargs)
+                               this_settings.metadata)
```
landrito: Why is this change happening? This seems like a breaking change.

lib/google/gax/settings.rb
```diff
@@ -62,20 +62,22 @@ class CallSettings
...
+                     metadata: {}, kwargs: {}, errors: [])
         @timeout = timeout
         @retry_options = retry_options
         @page_descriptor = page_descriptor
         @page_token = page_token
         @bundler = bundler
         @bundle_descriptor = bundle_descriptor
-        @kwargs = kwargs
+        @metadata = metadata
+        @metadata.merge!(kwargs) if kwargs && metadata
```
landrito: Might be good to add a deprecation warning to this.

lib/google/gax/api_callable.rb
geigerj: I think the `merge!` should avoid breaking current users, right?

Context on change: https://github.com/googleapis/gax-ruby/issues/86

apolcyn:
> Just for confirmation, @apolcyn -- is there any downside to our libraries setting return_op: true and blocking on the result for calls where we don't actually need the operation, as opposed to not setting return_op?

No, once the RPC has been started, the behavior of `return_op: true`, and `return_op: false` will be the same. `return_op` true effectively just allows splitting up instantiation and RPC invocation, and effectively just gives a different surface API.

lib/google/gax/settings.rb
landrito: I assume kwargs will in a manner with no time-frame of removal and removed at next major update, correct?

lib/google/gax/api_callable.rb
jbolinger: Yes, I kept the merge in to avoid breaking it.

lib/google/gax/settings.rb
jbolinger: Yup, is there a better way to add that note than what I've got?

landrito: You could use the [Kernel#warn](http://ruby-doc.org/core-2.2.0/Kernel.html#method-i-warn) method to simply emit a warning. We have done that before but I don't know if there is a more idiomatic way than that.

jbolinger:
@geigerj @landrito - not super urgent, but bumping so we can get this merged relatively soon.

Any outstanding concerns? The only issue I can recall is if we wanted to return nil when a block was passed to prevent exposing the same thing through the block and the return value. I don't really have an opinion either way. Thoughts? I miss anything else?

geigerj:
I was just waiting for a resolution to the question about on permitting both blocks and return values, and two other minor threads:

* Should we `warn` if users try to use the `kwargs` argument to pass headers? (IMO: yes, probably, but might be best to do this in a follow-up GAX release coordinated with google-cloud-ruby so that g-c-r users won't see warnings if they upgrade their dependencies GAX and g-c-r at the same time.)
* Is it possible to simplify the test mocking? (We chatted offline a little about this; it wasn't clear there was an easy way to.)

 I'm not blocking on a particular decision on these, as long as there is one.

jbolinger:
Let's leave the warning out for now since it's documented - we can add it later per your suggestion.

I haven't found a better way for the mocks. Happy to update if someone has a suggestion though. 

That just leaves us with the return issue. I don't have a strong opinion either way. I'm sort of inclined to leave it as-is simply because I found it useful for dev purposes (makes it easy to dump the server response to the console while doing other things in the block). Granted, that is a bit bizarre and I'm 100% willing to change it to return nil if there is any strong opinion to do so.

lib/google/gax/api_callable.rb
```diff
@@ -372,10 +380,17 @@ def retryable(a_func, retry_options, kwargs)
...
-        a_func.call(request, deadline: Time.now + timeout, metadata: kwargs)
+    def add_timeout_arg(a_func, timeout, metadata)
+      proc do |request, block|
+        op = a_func.call(request,
+                         deadline: Time.now + timeout,
+                         metadata: metadata,
+                         return_op: true)
+        res = op.execute
+        block.call op if block
+        res
```
landrito: A little bit confused that this works. If an `add_timeout` wraps a `retryable`, the `retryable` wrapping will cause the returned op to already call execute returning the result. If `add_timeout` does the same won't that throw an error?

lib/google/gax/api_callable.rb
geigerj: Ping on this.

lib/google/gax/api_callable.rb
jbolinger: Not sure I'm following, but sounds like it could be an issue. Could you help with an example or test case to clarify?

landrito: Lets say you have the method `foo` and you wrap it with retry and timeout like such:

`wrapped_foo = add_timeout( retryable(foo, ...) `

When `wrapped_foo` is called the call stack will look like:
```rb
foo_response = foo(request. return_op: true)
retryable_response = foo_response.execute
# this might error because execute has already been called to create the retryable response
add_timeout_timeout_response = retryable_response.execute
```

landrito: Furthermore, if there was a method wrapped by both retryable and add_timeout, the passed block will be called twice once as it is processed by retryable and once when it is processed by add_timeout.

landrito: Ah hm. I overlooked that the a method is either retryable or has timeout added so this might not be exactly a problem.

geigerj:
@jbolinger -- before you merge, could you check my comment on paginated calls here: https://github.com/googleapis/gax-ruby/pull/118/files#diff-30f305d8fc0cc5476762707ed876f556L229?

jbolinger:
Will do. I wanted to test that against a live API but haven't gotten to it. Can you think of a good one (not logging)?

geigerj:
@jbolinger you could try Pub/Sub. I think if you use the `gapic-test` project, there are already some test resources (i.e., subscriptions/topics/messages) set up for you to run against.

PR #127, opened 2018-05-25
Add interceptor support

jbolinger:
This change exposes gRPC interceptors at stub creation time.

@landrito for the corresponding generator change could we follow what you did for python? What would the Ruby code look like to add an interceptor to a client?

[will squash all the junk commits]


jbolinger:
Updated (also added channel_override since it was missing too).

Issue #126, opened 2018-05-04
PathTemplate#match returning different params that given to PathTemplate#render


blowmage:
While evaluating the behavior of `Google::Gax::PathTemplate`, I noticed a behavior that surprised me. Consider the following spec that I wrote to try to understand the existing behavior:

```ruby
it 'match returns the same params that render uses' do
  template = PathTemplate.new('v1/{name=parent/*}')
  params = { 'name' => 'parent/child' }
  path = 'v1/parent/child'

  expect(template.render(symbolize_keys(params))).to eq(path)

  expect(template.match(path)).to eq(params)
end
```

It fails with the following:

```
Failures:

  1) Google::Gax::PathTemplate match returns the same params that render uses
     Failure/Error: expect(template.match(path)).to eq(params)
     
       expected: {"name"=>"parent/child"}
            got: {"name"=>"child"}
     
       (compared using ==)
     
       Diff:
       @@ -1,2 +1,2 @@
       -"name" => "parent/child",
       +"name" => "child",
       
     # ./spec/google/gax/path_template_spec.rb:51:in `block (2 levels) in <top (required)>'
```

I find this surprising. Can someone verify that the behavior of `Google::Gax::PathTemplate#match` here is correct? If so, why isn't the value returned not matching the constraint given in the template?

pongad:
I don't think I'm the expert here, but I think the behavior here is definitely incorrect. The name should be "parent/child".

blowmage:
You will also note that I had to call `symbolize_keys` on the hash argument passed to `#render`, but the return value from `#match` was a hash with string keys. I don't understand why they are different. Shouldn't they either both use symbols keys or both use strings keys?

eoogbe:
@jbolinger Any updates?

jbolinger:
Path templates are being rewritten as part of https://github.com/googleapis/gapic-generator-ruby and this will be addressed in the next major release of this library.

blowmage:
I'm going to close this. The plan is to remove template handling in this gem.

Issue #119, opened 2018-02-26
Investigate Path template performance


geigerj:
Recent investigation by @blowmage found that GAPIC utility methods that rely on the GAX path template code are very slow, to the point that they are the bottleneck for some `google-cloud-ruby` acceptance tests. This is likely due to the use of `rly` to implement template parsing.

Investigate these performance concerns. If `rly` is the issue, `addressable` may be a good substitute, although be cautious of [differences between Google's path templates and RFC 6570](https://github.com/googleapis/googleapis/blob/ff51363884b7729ecf22481e0e9e136ea980d29e/google/api/http.proto#L253). Another approach might be to drop parsing and use basic string interpolation, which @pongad has already done in the Go GAPICs.

jonparrott:
Python also has just string interpolation (with a touch of validation). It's likely fairly trivial to port the Python implementation to Ruby: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/api_core/google/api_core/path_template.py

jonparrott:
Python also dropped the use of a full-blown parser and its path template module is extremely small, simple, and well tested: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/api_core/google/api_core/path_template.py

I don't think we should deprecated them, I think we should provide them as helpers to construct strings in the correct format.

jbolinger:
@jonparrott I'd bet you'd prefer that all things follow Python's approach though 😄 

I think we'll have a complication either way here. These are currently implemented as class methods (IIRC this was why @frankyn didn't like them), which I do think looks weird and often verbose. If we made them available as instance methods it would look exactly like Python, but we'd probably have to keep both around for awhile and it does introduce the (remote) possibility for things like name collisions.

I just have a hard time believing that folks really needs these things (i.e. in the pub sub docs only 3/7 languages use them). @blowmage do you think we'll get complaints if we don't have these, or have we gotten issues/complaints related to this in the past?

Let's also check with our fellow Python friend @lukesneeringer - he had a long conversation about this the other day and I'd be curious what he'd suggest we do.

jonparrott:
> @jonparrott I'd bet you'd prefer that all things follow Python's approach though 😄

Bias aside, the Python implementation was well-thought out and intentionally intended to provide a basis for other languages and was reviewed by several others (including @lukesneeringer). This actually goes for *all* of api core.

[...]

PR #128, opened 2018-05-30
Path template potential re-write

jbolinger:
This is an example of a potential zero dependency replacement for the path template helpers to address the problems noted in #119 and #126 (excluding the inconsistency of `symbolize_keys` comment). It drops the need for rly and replaces the parsing logic with regular expressions.

@blowmage PTAL whenever you get a chance to verify if this change would actually be useful to you. I'm not too familiar with the tests it was impacting, or I might have tried to benchmark it myself, but if this helps we can find a way to get it in.  Consider it a WIP for now.

DO NOT MERGE: the API is nearly identical but we may need to keep the original implementation around or include both (TBD).

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -13,7 +13,7 @@ AllCops:
 
 # Offense count: 1
 Metrics/AbcSize:
-  Max: 38
+  Max: 44
 
 # Offense count: 1
 Metrics/CyclomaticComplexity:
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -29,6 +29,9 @@ Metrics/LineLength:
 Metrics/MethodLength:
   Max: 40
 
+Metrics/ClassLength:
+  Max: 250
+
 # Offense count: 1
 Metrics/PerceivedComplexity:
   Max: 11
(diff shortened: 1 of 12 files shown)


jbolinger:
Yeah, that doesn't surprise me. If you send me the tests I can try to update and see how far I can get it. 

For backwards compatibility we've got options. I was thinking the simplest path might be to just include both implementations in the library for awhile and update any references in google-cloud-ruby to use the newer one to address the performance issues. I doubt this gets much use beyond that repo and given that this isn't a whole lot of code the duplication isn't too bad. Replacing would be cleaner but I just don't know if we've got the right coverage to do it with confidence.

jbolinger:
On that same note, let me know how much of a difference this actually makes for you. I don't recall how significant the overhead was and it got labeled as a relatively low priority work item. If it's causing a significant amount of degradation across APIs I suspect our devrel folks would be more open to potentially disruptive changes.

blowmage:
From our usage via the GAPIC layer, we won't see any impact from the behavior change. But if someone is using this class directly (not using the GAPIC layer), they may be impacted by the change.

I've been trying to get a performance benchmark all morning, it just hasn't happened yet.

jbolinger:
Closing since the plan is to change this in the next major release.

PR #134, opened 2018-08-16
Lookup Operation types to unpack

blowmage:
This PR changes `Google::Gax::Operation` to lookup the result and metadata types used to unpack the values from the `Google::LongRunning::Operation` object. This only happens when the `result_type` or `metadata_type` values are passed in as `nil` (which currently results in an error).

This change will enable the GAPIC client objects to retrieving operations through a `get_operation` and a `list_operation` method without having to know the specific result and metadata types to set. There are several APIs that have operation collections which contain resources with varied result and metadata types. Each `Google::LongRunning::Operation` object would have to be inspected and the result and metadata types discovered before passing to `Google::Gax::Operation.new`. This PR moves the result and metadata types discovery to `Google::Gax::Operation`.

### DIFF
--- a/lib/google/gax/operation.rb
+++ b/lib/google/gax/operation.rb
@@ -33,7 +33,6 @@ require 'time'
 require 'google/gax/constants'
 require 'google/gax/settings'
 require 'google/protobuf/well_known_types'
-require 'uri'
 
 module Google
   module Gax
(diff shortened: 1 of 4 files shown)


lib/google/gax/operation.rb
```diff
@@ -155,7 +156,19 @@ def name
       # @return [Object, nil]
       #   The metadata of the operation. Can be nil.
       def metadata
-        return nil if @grpc_op.metadata.nil?
+        return if @grpc_op.metadata.nil?
+
+        @metadata_type ||= begin
+          lookup_uri = URI.parse(@grpc_op.metadata.type_url)
```
jbolinger: Is this necessary? Will `grpc_op.metadata.type_name` get you the same result without the URI parsing? Ex:

```
require 'google/cloud/speech'

ctx = Google::Cloud::Speech::V1::SpeechContext.new
ctx.phrases << "a" << "b"

any = Google::Protobuf::Any.new
any.pack ctx

type = Google::Protobuf::DescriptorPool.generated_pool.lookup(any.type_name).msgclass
obj = any.unpack type
```

blowmage: I'll update this PR to use `#type_name`.

blowmage:
With this change, the GAPIC generators could be changed to specify `nil` to activate the lookup everytime `#metadata` or `#response` is called.

```ruby
def long_running_recognize \
    config,
    audio,
    options: nil
  req = {
    config: config,
    audio: audio
  }.delete_if { |_, v| v.nil? }
  req = Google::Gax::to_proto(req, Google::Cloud::Speech::V1::LongRunningRecognizeRequest)
  operation = Google::Gax::Operation.new(
    @long_running_recognize.call(req, options),
    @operations_client,
    nil, # was Google::Cloud::Speech::V1::LongRunningRecognizeResponse
    nil, # was Google::Cloud::Speech::V1::LongRunningRecognizeMetadata
    call_options: options
  )
  operation.on_done { |operation| yield(operation) } if block_given?
  operation
end
```

PR #136, opened 2018-09-25
Use descriptor pool for deserialization

jbolinger:
This library attempts to automatically `unpack` the error messages details field (an `Any` proto) when an RPC fails. This was originally implemented by reconstructing the expected Ruby class name from the `type_name` property of the `Any` object. 

This change replaces that by `unpack` ing the object using the generated descriptor pool instead. A `ruby_namespace` option was added to protobuf around version 3.6, which makes the original implementation unreliable when used.

### DIFF
--- a/lib/google/gax/grpc.rb
+++ b/lib/google/gax/grpc.rb
@@ -65,12 +65,13 @@ module Google
             error.to_status
           ).details
         details.map do |any|
-          # If the type of the proto wrapped by the Any instance is not
-          # available, do not deserialize.
-          candidate_class_name = class_case(any.type_name.split('.')).join('::')
+          # deserialize the proto wrapped by the Any in the error details
           begin
-            any.unpack(Object.const_get(candidate_class_name))
-          rescue NameError
+            type = Google::Protobuf::DescriptorPool.generated_pool.lookup(
+              any.type_name
+            )
+            any.unpack(type.msgclass)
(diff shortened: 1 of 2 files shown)


jbolinger:
@blowmage @dazuma have any of the cloud libraries been released using the `ruby_namespace` option?

lib/google/gax/grpc.rb
```diff
@@ -65,12 +65,13 @@ def deserialize_error_status_details(error)
...
-          # available, do not deserialize.
-          candidate_class_name = class_case(any.type_name.split('.')).join('::')
+          # deserialize the proto wrapped by the Any in the error details
           begin
-            any.unpack(Object.const_get(candidate_class_name))
-          rescue NameError
+            type = Google::Protobuf::DescriptorPool.generated_pool.lookup(
+              any.type_name
+            )
+            any.unpack(type.msgclass)
```
blowmage: It is possible for `type` to be nil, if the descriptor hasn't been loaded, AFAIK. We may want to guard against that.

blowmage:
I am not aware of any libraries making use of the `ruby_namespace` feature.

lib/google/gax/grpc.rb
blowmage: Ugh, I missed the notification on this. That works.

PR #138, opened 2018-10-24
Update googleauth dependency

blowmage:
Allow the dependency to use the 0.7.x releases.

### DIFF
--- a/google-gax.gemspec
+++ b/google-gax.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.0.0'
 
-  s.add_dependency 'googleauth', '~> 0.6.2'
+  s.add_dependency 'googleauth', '>= 0.6.2', '< 0.8.0'
   s.add_dependency 'grpc', '>= 1.7.2', '< 2.0'
   s.add_dependency 'googleapis-common-protos', '>= 1.3.5', '< 2.0'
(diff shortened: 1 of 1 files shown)


jbolinger:
@landrito do you know if this version was pinned for any reason?

@blowmage I assume that you've already tested this with the cloud apis?

landrito:
I remember no reason why it was pinned so strictly. Sorry! LGTM assuming some hand tests show nothing breaks.

blowmage:
> @blowmage I assume that you've already tested this with the cloud apis?

Not yet. This PR is part of incorporating some new functionality released in the latest version of googleauth into the cloud gems.

jbolinger:
@blowmage have you been able to test it with this updated auth version?

blowmage:
No, I have not. Isn't making the change and running this repo's existing CI sufficient?

jbolinger:
That's sufficient to merge but I don't think the existing CI runs any real tests so I would like to verify before releasing it. Do you need one soon?

blowmage:
I've tested this against google-cloud-ruby with googleauth 0.7.1 and everything works as expected.

blowmage:
I noticed this PR was out of date. I have updated it to the latest `master`. Can someone take a look?

jbolinger:
Thanks for verifying with google-cloud-ruby @blowmage! I'll merge it now.

PR #141, opened 2018-12-03
Update Operations documentation

### LINKED ISSUE #139: Update Operation docs
Document that [`result_type` and `metadata_type`](https://github.com/googleapis/gax-ruby/blob/master/lib/google/gax/operation.rb#L108-L111) may be `nil` thanks to #134.

blowmage:
* Document lookup behavior of `result_type` and `metadata_type` type classes when passed in as `nil`.
* Document `call_options` as an optional argument.

[closes #139]

### DIFF
--- a/lib/google/gax/operation.rb
+++ b/lib/google/gax/operation.rb
@@ -105,9 +105,9 @@ module Google
       #   The inital longrunning operation.
       # @param client [Google::Longrunning::OperationsClient]
       #   The client that handles the grpc operations.
-      # @param result_type [Class] The class type to be unpacked from the
+      # @param result_type [Class, nil] The class type to be unpacked from the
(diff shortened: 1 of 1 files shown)


lib/google/gax/operation.rb
```diff
@@ -106,11 +106,12 @@ class Operation
       # @param client [Google::Longrunning::OperationsClient]
       #   The client that handles the grpc operations.
       # @param result_type [Class] The class type to be unpacked from the
```
quartzmo: Want to change the types list to `[Class, nil]` for these?

blowmage: Updated.

quartzmo (approved):

PR #142, opened 2018-12-03
Make Operation class type arguments optional

blowmage:
This PR changes the `Google::Gax::Operation` `result_type` and `metadata_type` arguments to be optional. This is a backwards compatible change.


blowmage:
@jbolinger @quartzmo Thoughts?

jbolinger:
Relying on the registry to do the lookup is probably the common case, so adding the `nil` defaults seems like a good idea to me.

PR #143, opened 2019-01-03
Loosen googleauth dependency

blowmage:
Loosen the googleauth dependency to allow future changes. The team supporting googleauth has committed to not change backwards compatibility before 0.10.0.


blowmage:
@jbolinger @landrito Ping.

jbolinger:
Ok, I will release a new version shortly.

PR #160, opened 2019-04-10
Remove RPC retry

blowmage:
Retry will be restored with a different implementation in the future.


blowmage:
Oops, this was already submitted and merged in #157.

PR #161, opened 2019-04-10
Remove ClientConfig

### LINKED ISSUE #149: Remove JSON Configuration
The next release will remove the JSON configuration, and all code related to the JSON configuration.

blowmage:
[closes #149]


lib/google/longrunning/operations_client.rb
```diff
@@ -347,6 +318,24 @@ def delete_operation \
...
+        package_version = Gem.loaded_specs['google-gax'].version.version
+
+        google_api_client = ["gl-ruby/#{RUBY_VERSION}"]
+        google_api_client << "#{lib_name}/#{lib_version}" if lib_name
+        google_api_client << "gapic/#{package_version}"
+        google_api_client << "gax/#{Google::Gax::VERSION}"
+        google_api_client << "grpc/#{GRPC::VERSION}"
+        google_api_client.join " "
+
+        headers = { 'x-goog-api-client' => google_api_client }
```
quartzmo: No longer need to freeze the `google_api_client` string? Will frozen strings be default from the supported Ruby versions?

lib/google/longrunning/operations_client.rb
```diff
@@ -347,6 +318,24 @@ def delete_operation \
...
+        headers.merge! metadata unless metadata.nil?
+
+        Google::Gax.const_get(:CallSettings).new metadata: headers
```
quartzmo: So we no longer need to set `timeout`, `page_descriptor`, or `errors` as was previously done in `construct_settings`?

blowmage: Correct. These values will either be passed in differently, or will be removed from 2.0.

lib/google/longrunning/operations_client.rb
blowmage: Correct. A future change will add google-style and the magic frozen string comment.

PR #164, opened 2019-04-11
Remove CallSettings

### LINKED ISSUE #148: Remove CallSettings
The next release will remove CallSettings and instead use a combination of sensible defaults and error introspection for this functionality.

blowmage:
[closes #148]

### DIFF
--- a/lib/google/gax/api_callable.rb
+++ b/lib/google/gax/api_callable.rb
@@ -47,31 +47,42 @@ module Google
     # same signature as the original.
     #
     # @param func [Proc] used to make a bare rpc call
-    # @param settings [CallSettings] provides the settings for this call
+    # @param timeout [Numeric] client-side timeout for API calls
+    # @param metadata [Hash]  request headers
     # @param params_extractor [Proc] extracts routing header params from the
     #   request
     # @param exception_transformer [Proc] if an API exception occurs this
     #   transformer is given the original exception for custom processing
     #   instead of raising the error directly
-    # @return [Proc] a bound method on a request stub used to make an rpc call
-    def create_api_call(func, settings, params_extractor: nil,
-                        exception_transformer: nil)
-      api_caller = proc do |api_call, request, _settings, block|
-        api_call.call(request, block)
-      end
+    # @return [Proc] A bound method on a request stub used to make an rpc call.
+    #   The argumentd for the bound method are:
+    #
+    #   * request - Request object
+    #   * options - CallOptions object
+    #   * block named argument - Proc object for yielding the operation
+    def create_api_call(func, timeout: 60, metadata: {},
+                        params_extractor: nil, exception_transformer: nil)
+      proc do |request, options = nil, block = nil|
+        options = CallOptions.new if options.nil?
 
-      proc do |request, options = nil, &block|
-        this_settings = settings.merge(options)
+        options.timeout  = timeout  if options.timeout  == :OPTION_INHERIT
+        options.metadata = metadata if options.metadata == :OPTION_INHERIT
+
         if params_extractor
           params = params_extractor.call(request)
-          this_settings = with_routing_header(this_settings, params)
+          routing_header = params.map { |k, v| "#{k}=#{v}" }.join('&')
+          options.metadata['x-goog-request-params'] = routing_header
         end
-        api_call = add_timeout_arg(func, this_settings.timeout,
-                                   this_settings.metadata)
+
         begin
-          api_caller.call(api_call, request, this_settings, block)
-        rescue *settings.errors => e
-          error_class = Google::Gax.from_error(e)
+          op = func.call(request, deadline: Time.now + options.timeout,
+                                  metadata: options.metadata,
+                                  return_op: true)
+          res = op.execute
+          block.call op if block
+          res
+        rescue GRPC::BadStatus => grpc_error
+          error_class = Google::Gax.from_error(grpc_error)
           error = error_class.new('RPC failed')
           raise error if exception_transformer.nil?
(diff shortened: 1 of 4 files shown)


lib/google/gax/api_callable.rb
```diff
@@ -132,8 +132,6 @@ def initialize(request_page_token_field,
       #   A proc to update the response object.
       # @param request [Object]
       #   The initial request object.
-      # @param settings [CallSettings]
-      #   The call settings to enumerate pages
       # @return [PagedEnumerable]
       #   returning self for further uses.
       def start(api_call, request, settings, block)
```
quartzmo: Should `settings` still be in this method signature?

blowmage: Yeah, it might be a good idea to merge #166 before this PR.

blowmage:
> settings is still mentioned in the create_api_call docs.

Good catch. Updated the docs

lib/google/gax/api_callable.rb
quartzmo: OK, I'll leave that up to you.

blowmage:
@jbolinger @quartzmo This PR is ready to be reviewed now.

lib/google/gax/api_callable.rb
blowmage: This has been fixed.

lib/google/gax/api_callable.rb
```diff
@@ -33,45 +33,54 @@
...
-      api_caller = proc do |api_call, request, _settings, block|
-        api_call.call(request, block)
-      end
+    # @return [Proc] A bound method on a request stub used to make an rpc call.
+    #   The argumentd for the bound method are:
+    #
+    #   * request - Request object
+    #   * options - CallOptions object
+    #   * block named argument - Proc object for yielding the operation
+    def create_api_call(func, timeout: 60, metadata: {},
```
blowmage: Instead of passing a CallSettings object, the values are passed in individually. This means that if a value is not passed in, a default value will be used instead.

lib/google/gax/api_callable.rb
```diff
@@ -33,45 +33,54 @@
...
-    # which function decorators to apply.
+    # defined in this module to +func+.
     #
-    # The result is another proc which for most values of +settings+ has the
-    # same signature as the original.
+    # The result is another proc which has the same signature as the original.
     #
     # @param func [Proc] used to make a bare rpc call
-    # @param settings [CallSettings] provides the settings for this call
+    # @param timeout [Numeric] client-side timeout for API calls
```
jbolinger: is this in seconds?

lib/google/gax/api_callable.rb
```diff
@@ -33,45 +33,54 @@
...
+    # @param metadata [Hash]  request headers
     # @param params_extractor [Proc] extracts routing header params from the
```
jbolinger: Just a thought (and I'll leave it up to you to decide)... since we're changing a bunch of parameters on this method already should we consider generalizing this one? 

When I read this as-is it's not very clear what this parameter is for and what a routing header is. One option might be to change this to a more generic proc that takes the request and returns a hash to merge with the request headers. This would also prevent us from embedding `x-goog-request-params` by default and we could move that reference into the generator.

jbolinger: Or, we might just remove it entirely since you've pulled `metadata` into the parameter list.

lib/google/longrunning/operations_client.rb
```diff
@@ -124,7 +124,8 @@ def initialize \
           updater_proc = credentials.updater_proc
         end
 
-        defaults = default_settings(metadata, lib_name, lib_version)
+        _default_timeout = default_timeout
```
jbolinger: nit: why not just call this something else rather than use `_`?

lib/google/gax/api_callable.rb
blowmage: Yes, it is seconds. I want to do a full documentation pass eventually.

lib/google/gax/api_callable.rb
blowmage: I am actually leaning towards having more of these procs that do interesting things to the request. We will add another proc for streaming responses, and I would like to wrap responses in `Google::Gax::Operation` or `PagedEnumerable` by providing a proc that does this, rather than by doing in in the client method.

jbolinger: Ah, a proc for transforming the output does sounds useful. Let's do that in a new change then and address them all.

lib/google/longrunning/operations_client.rb
blowmage: Oh yeah, I have this different in other working branches. I'll rename this. I think `default_timeout` goes away, since it should use `timeout` here anyways.

blowmage: Update.

lib/google/gax/api_callable.rb
blowmage: Yes. I have many, many branches with different changes ready to be posted in new PRs. :)

blowmage:
> It would admittedly, make this an even more drastic change and shift some responsibility to the generator.

This is where I am leaning. The more we can put in the generator, the less that the runtime has to manage. The PathTemplate class is a great example. We can parse the resource path templates in the generator and output a simple ruby module that has no other runtime dependencies. I think there are opportunities for more changes like that across the microgenerator and gax 2.0

lib/google/gax/api_callable.rb
jbolinger: Ok, we can circle back then.

PR #166, opened 2019-04-11
Refactor PagedEnumerable

blowmage:
Update PagedEnumerable to work by convention, not configuration.
Paged API calls will be determined by fields existing on the
request object and the response object.
Remove configuration and simplify the way PagedEnumerable objects
are created. We expect the GAPIC client to create instead of GAX.


lib/google/gax/paged_enumerable.rb
```diff
@@ -0,0 +1,200 @@
...
+      #   The name of the field in the response which holds the next page token.
+      # @param resource_field [String]
+      #   The name of the field in the response which holds the resources.
+      def initialize(api_call, request, response, options: nil,
+                     resource_field: nil)
+        @api_call = api_call
+        @request = request
+        @response = response
+        @options = options
+        @resource_field = resource_field || determine_resource_field(response)
```
jbolinger: Do we need this default vs raising if the `resource_field` is not specified? (Also need to update the method docs)

lib/google/gax/paged_enumerable.rb
```diff
@@ -0,0 +1,200 @@
...
+        @page.response
+      end
+
+      private
+
+      # This inspects a response object and pulls out the first repeated field.
+      def determine_resource_field(response)
+        # Find the first repeated FieldDescriptor on the response Descriptor
+        # Find uses the order of the FieldDescriptor on the Descriptor
+        field = response.class.descriptor.find { |f| f.label == :repeated }
```
jbolinger: worth putting a similar sanity check in the initializer so we know the field exists and is of the correct format?

lib/google/gax/paged_enumerable.rb
```diff
@@ -0,0 +1,200 @@
...
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+module Google
+  module Gax
+    # A class to provide the Enumerable interface for page-streaming method.
```
jbolinger: I know this text was copied, but what do you think about the term `page-streaming`? I don't think it's a big deal but calling it a `paginated method`, or anything without the term `streaming` may be better so it's clear this has nothing to do with grpc streaming.

blowmage:
> If we go ahead without requiring the response field to be explicitly given do we need to update the tests? Any others?

Funny enough, there are zero tests for PagedEnumerable. So yes, we need to update/add tests. We had discussed migrating from RSpec to Minitest, and I would prefer to add tests using Minitest. I'll add an issue to track the missing tests for PagedEnumerable.

lib/google/gax/paged_enumerable.rb
blowmage: I could go either way. We will probably be adding the logic to determine the first repeated field in the generator. But, since paging is supposed to be implemented by convention I had also thought that the convention could be followed in the runtime if the field name wasn't provided.

I had also wondered if we want to check the `request` object for `page_token` and `page_size`, and the `response` object for `next_page_token` and a repeated field, per the convention. That could possibly save the need for passing a resource_field, and ensure that anyone using GAX outside of the new generator are using it correctly.

What do you think?

lib/google/gax/paged_enumerable.rb
blowmage: Yeah, this is partially why I can see us not passing the `resource_field` at all, and always looking it up.

lib/google/gax/paged_enumerable.rb
blowmage: Agreed. I updated the documentation. What do you think?

jbolinger: Looks good.

lib/google/gax/paged_enumerable.rb
jbolinger: That sounds fine to me. In that case, I think we may want to make tighten the check in `#determine_resource_field` to ensure that there is exactly one `:repeated` field when it's inferred.

blowmage: Ah, that was part of what I was trying to figure out earlier this week. The unofficial word is that we take the first repeated field, but are you saying that there can only be one repeated field in a paginated response? Is that the official convention?

jbolinger: The "official" convention is IMHO a bit convoluted:

> If the response message has more than one non-primitive repeated field, the first one (in order of appearance in the file and field number) is used. If the first field by order of appearance in the message and the first field by field number do not match, code generators that implement automatic pagination should fail with an error.

But, we don't really have to implement this to the letter. If we go with this as-is then we also need to check the field number to be consistent with the convention. However, I think we should simply require the user to explicitly state the field that they want to use if there's more than one. That's easier to explain than the field number check and it will make more sense if gax is used outside of the generator.

blowmage: Oh, are primitive fields allowed to be paginated then?

jbolinger: No, primitive fields will not be paged by convention. We could allow primitives to be paged in gax if we really wanted to, but we'll never use it in the generator, so I don't think it's worth worrying about unless we get an external ask for it.

blowmage: K, I have added a commit that attempts to implement these rules. There are no tests yet, but they will come later. There are a few more changes I would like to make before backfilling the tests. But they are coming.

lib/google/gax/paged_enumerable.rb
```diff
@@ -136,17 +138,56 @@ def response
...
+        end
+        if page_token.nil?
+          raise ArgumentError.new(
+            'request must have a page_token field (String)'
+          )
+        end
+
+        page_size = @request.class.descriptor.find do |f|
+          # Google::Protobuf::FieldDescriptorProto::Type::TYPE_INT32 = 5
+          f.name == 'page_size' && f.type == 5
```
jbolinger: We should probably allow all the integer types here and not just `TYPE_INT32`.

blowmage: The design doc says:

> Pagination can be inferred when the input message for an RPC contains an `int32` `page_size` field and a `string` `page_token` field.

jbolinger: True, but I can assure you that it's over-constrained and I don't see any harm in relaxing it. It has slipped through at least a couple of times: https://github.com/googleapis/googleapis/blob/738b3fa5915ea6000fe5a3054e12bbf8db681662/google/cloud/support/v1alpha1/cloud_support.proto#L163

blowmage: So allow any of the integer types then?

```
TYPE_INT64 = 3
TYPE_UINT64 = 4
TYPE_INT32 = 5
TYPE_FIXED64 = 6
TYPE_FIXED32 = 7
TYPE_UINT32 = 13
TYPE_SFIXED32 = 15
TYPE_SFIXED64 = 16
TYPE_SINT32 = 17
TYPE_SINT64 = 18
```

blowmage: Or limit it to just `TYPE_INT64` and `TYPE_INT32` (3, 5)?

blowmage: @jbolinger I have added int64. Do you want the additional 8 types added as well?

jbolinger: The others are much less common. I think these two are fine for now.

PR #168, opened 2019-04-11
Refactor BackoffSettings

blowmage:
* Make BackoffSettings a class with optional named arguments for the constructor, instead of a Struct with positional arguments without a default value.
* This will allow users to modify only the values they care about when creating a new BackoffSettings object.
* Change BackoffSettings to use seconds instead of milliseconds.
* Remove `MILLIS_PER_SECOND` as it is no longer needed.
* Remove `google/gax/constants.rb` file.


blowmage:
BackoffSettings isn't used in retry AFAIK. I think it is only used in `Operation#wait_until_done!`.

blowmage:
Oh, you are right. They were included in RetryOptions before it was removed. I can sit on this change until we implement the updated error retry.

blowmage:
Adding `do not merge` until #166 is approved and merged. This PR will need to be rebased to fix conflicts.

blowmage:
This is satisfied by #178.

Issue #165, opened 2019-04-11
More visibility/control in error retry


blowmage:
See the [discussion in #150](https://github.com/googleapis/gax-ruby/issues/150#issuecomment-482272294) for an example of the desired visibility/control.

blowmage:
As discussed in #172, we are considering delegating the retry and backoff calculations to a Retry object, and allowing users to provide their own object to have more control on errors and retry behavior. We would need to design a simple interface that was easily reproduced for users to create, if they chose not to simply configure the default Retry object.

blowmage:
@jbolinger Do you think this issue is satisfied by #172?

Issue #150, opened 2019-04-05
Rewrite retry mechanism


blowmage:
Rewrite retry mechanism by inspecting raised error only, no configuration.

leavingonaspaceship:
@blowmage Can you provide some more details on what the plan is here? We use `gax-ruby` via `google-cloud-spanner` and have disabled the `gax-ruby` retry settings and handle retries and timeouts on our end. Will that still be possible with the current plan to not allow retry configuration in `gax-ruby`?

jbolinger:
@leavingonaspaceship the plan is to remove the existing retry configuration and replace it with something simpler that gives the library users more direct control on how retrying works, if they want to use automatic retry at all. We'll definitely continue to allow it to be disabled. 

If you think there's anything we could do better to support your particular use case, beyond disabling it, let us know and we'll see how we can help.

leavingonaspaceship:
Gotcha. Does that mean we should be looking for changes in how `google-cloud-spanner` handles retrying after `gax-ruby@2.0` is released?

The main problem we had with the way retrying is currently handled is that we couldn't instrument it at all. We couldn't tell the difference between a single 1-second latency vs. 4 0.25-second latencies. Errors like DeadlineExceededError were also swallowed (except in particularly bad cases of 60-second requests), so we couldn't see how often different errors were happening.

blowmage:
It means that you will likely see changes to configuration in the future. Hopefully it will be an improvement. I don't think anyone know for sure if clients such as google-cloud-spanner will be 100% backwards compatible when on gax 2.0, so they may see a version bump as well. (Clients like Spanner are a challenge because they contain a public UI, but also the low-level generated code to access the API that is also public. So any change to the low level API code is right now technically public, and won't be backwards compatible when switching to gax 2.0.)

blowmage:
@jbolinger Do you think this issue is satisfied by #172?

jbolinger:
@blowmage yes.

PR #172, opened 2019-04-12
Reimplement retry with incremental backoff

### LINKED ISSUE #163: Merge kwargs and metadata in CallOptions
`kwargs` was renamed to `metadata` since it was used exclusively for metadata, but the existing parameter was kept around to avoid a breaking change. `kwargs` should be dropped.

blowmage:
This implements the simplified retry mechanics. It splits apart the old RetryOptions object and places `retry_codes` and `retry_settings` directly on `CallOptions`. This was done to allow users to configure only the settings they want to change, without having to specify all settings.

RetrySettings is a replacement for BackoffSettings. It provides a simpler set of values to configure the incremental backoff. It also allows users to only specify the values they want to change, not every value.

CallOptions#kwargs has been removed.

[refs #150, fixes #163, refs #165]


lib/google/longrunning/operations_client.rb
```diff
@@ -141,19 +141,23 @@ def initialize \
 
         @get_operation = Google::Gax.create_api_call(
           @operations_stub.method(:get_operation),
-          timeout: timeout, metadata: metadata
+          timeout: timeout, metadata: metadata,
+          retry_codes: [14]
```
jbolinger: Aren't these codes defined on `GRPC::Core::StatusCodes` (i.e. `::CANCELLED`, etc.)?

lib/google/gax/api_callable.rb
```diff
@@ -47,6 +50,8 @@ module Gax
     # @param func [Proc] used to make a bare rpc call
     # @param timeout [Numeric] client-side timeout for API calls
     # @param metadata [Hash]  request headers
+    # @param retry_codes [Array<Integer>] The error codes to retry
```
jbolinger: Instead of passing these `retry_*` arguments as parameters here and forwarding them to a new `RetryManager` why not pass a `RetryManager` here directly and allow the user to customize via subclassing or by creating a new instance of the default `RetryManager` with non-default arguments?

lib/google/gax/api_callable.rb
```diff
@@ -58,33 +63,41 @@ module Gax
...
           res
-        rescue GRPC::BadStatus => grpc_error
-          error_class = Google::Gax.from_error(grpc_error)
-          error = error_class.new('RPC failed')
-          raise error if exception_transformer.nil?
-          exception_transformer.call error
         rescue StandardError => error
+          if retry_manager.retry?(error)
+            retry_manager.delay!
+            retry
```
jbolinger: What is `retry` here?

lib/google/gax/call_settings.rb
```diff
@@ -0,0 +1,84 @@
...
+        @retry_codes ||= retry_codes
+        @retry_settings ||= RetrySettings.from(retry_settings) if retry_settings
+      end
+
+      # @private
+      def set_default_values_for_internal_use!
+        # Set defaults if missing
+        @timeout        ||= 300 # 5 minutes
+        @metadata       ||= {}
+        @retry_codes    ||= [14] # always retry on UNAVAILABLE
```
jbolinger: Replace with constant (see previous comment).

lib/google/gax/errors.rb
```diff
@@ -143,7 +143,7 @@ def self.grpc_error_class_for(grpc_error_code)
        PermissionDeniedError, ResourceExhaustedError, FailedPreconditionError,
        AbortedError, OutOfRangeError, UnimplementedError, InternalError,
        UnavailableError, DataLossError,
-       UnauthenticatedError][grpc_error_code] || GaxError
+       UnauthenticatedError][grpc_error_code.to_i] || GaxError
```
jbolinger: why did this change?

lib/google/gax/retry_manager.rb
```diff
@@ -0,0 +1,72 @@
...
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+require 'google/gax/retry_settings'
+
+module Google
+  module Gax
+    class RetryManager
```
jbolinger: I think we should document this as a public interface and document how to customize it here (see previous comment).

lib/google/gax/retry_settings.rb
```diff
@@ -0,0 +1,89 @@
...
+    # @!attribute delay_multiplier
+    #   @return [Numeric] the multiplier by which to increase the
+    #     delay time between the completion of failed requests, and
+    #     the initiation of the subsequent retrying request.
+    # @!attribute max_delay
+    #   @return [Numeric] the maximum delay time, in seconds,
+    #     between requests. When this value is reached,
+    #     +delay_multiplier+ will no longer be used to
+    #     increase delay time.
+    class RetrySettings
```
jbolinger: I'm not sure that we really need this class. These could be hash args that are passed to `RetryManager#initialize`, which might make it a little cleaner for custom retry implementations that are not using a timed backoff approach.

lib/google/longrunning/operations_client.rb
```diff
@@ -141,19 +141,23 @@ def initialize \
...
         )
         @list_operations = Google::Gax.create_api_call(
           @operations_stub.method(:list_operations),
-          timeout: timeout, metadata: metadata
+          timeout: timeout, metadata: metadata,
+          retry_codes: [14]
```
jbolinger: Use constants (see previous comment)?

lib/google/longrunning/operations_client.rb
```diff
@@ -141,19 +141,23 @@ def initialize \
...
         )
         @cancel_operation = Google::Gax.create_api_call(
           @operations_stub.method(:cancel_operation),
-          timeout: timeout, metadata: metadata
+          timeout: timeout, metadata: metadata,
+          retry_codes: [14]
```
jbolinger: Use constants (see previous comment)?

lib/google/longrunning/operations_client.rb
```diff
@@ -141,19 +141,23 @@ def initialize \
...
         )
         @delete_operation = Google::Gax.create_api_call(
           @operations_stub.method(:delete_operation),
-          timeout: timeout, metadata: metadata
+          timeout: timeout, metadata: metadata,
+          retry_codes: [14]
```
jbolinger: Use constants (see previous comment)?

lib/google/gax/call_settings.rb
```diff
@@ -0,0 +1,84 @@
...
+    # Encapsulates the overridable settings for a particular API call
+    # @!attribute [rw] timeout
+    #   @return [Numeric, nil]
+    # @!attribute [rw] metadata
+    #   @return [Hash, nil]
+    # @!attribute [rw] retry_codes
+    #   @return [Array<Integer>, nil]
+    # @!attribute [rw] retry_settings
+    #   @return [RetrySettings, nil]
+    class CallOptions
```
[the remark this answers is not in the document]
blowmage: Because I was up until 4:00 am working on this stuff. Fixed.

lib/google/longrunning/operations_client.rb
blowmage: Yes, changed.

lib/google/gax/call_settings.rb
blowmage: Changed.

lib/google/gax/api_callable.rb
blowmage: It replays the block, starting at the `begin` a few lines up. This is how the API call is made again.

lib/google/gax/call_settings.rb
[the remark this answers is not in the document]
jbolinger: 🤣 well, believe me, it was more fun than what I was doing around 4:00 am yesterday!

lib/google/gax/api_callable.rb
blowmage: Ohhh, so give users the ability to completely control the retry mechanics, even user their own manager object if they want to? I like it! So object oriented!

We will still need a way to configure different client methods with different retry configuration. But I think we can do something here that will better address #165. Do you want that to be done in this PR or do you want that to be a follow up PR?

lib/google/gax/api_callable.rb
jbolinger: doh, keyword gotcha.

lib/google/gax/retry_settings.rb
blowmage: I agree that we can do better. retry codes, RetrySettings, and RetryManager can probably be squashed into one object.

lib/google/gax/errors.rb
blowmage: I had some test error object that had a code method, but it returned a non-integer value. So this is a bit of protection.

lib/google/gax/retry_manager.rb
blowmage: Agreed. I'll think about this over the weekend.

lib/google/gax/retry_settings.rb
blowmage: I've thought about this, and I think I have a design that will collapse RetrySettings and RetryManager into a single object. If a user provides a Hash then we will create this new object, or the user can pass an object that implements `#retry?(error)` and `#delay!` methods instead. That should also satisfy giving users more visibility/control of the errors raised and the retry mechanics. Working on this today.

lib/google/gax/api_callable.rb
blowmage: I have added RetryPolicy, which can be configured. Or, users can provide a Proc (or any object that implements a `#call` method) to determine if the call should be retried, and implement the incremental backoff. PTAL.

lib/google/gax/retry_manager.rb
blowmage: I've changed this and added documentation. Please take a look and let me know what you think.

lib/google/gax/retry_settings.rb
blowmage: This class has been removed.

lib/google/gax/api_call.rb
```diff
@@ -0,0 +1,136 @@
...
+      # used to make an RPC call.
+      #
+      # The result is created by applying a series of function decorators
+      # defined in this module to `stub_method`.
+      #
+      # The result is another proc which has the same signature as the original.
+      #
+      # @param stub_method [Proc] used to make a bare rpc call
+      # @param timeout [Numeric] client-side timeout for API calls
+      # @param metadata [Hash]  request headers
```
jbolinger: nit: extra space?

lib/google/gax/api_call.rb
```diff
@@ -0,0 +1,136 @@
...
+      # @param retry_policy [Hash] the settings for error retry, will be merged
+      #   to the {CallOptions#retry_policy} object if supported.
+      # @param params_extractor [Proc] extracts routing header params from the
```
jbolinger: I mentioned this earlier, but I'm still in favor of making this more generic and pushing the routing headers into the generator.

For example, instead of a `params_extractor` that specifically deals with `x-goog-*` parameters could this be a proc that given the request and options returns a hash of headers to be sent?

lib/google/gax/call_options.rb
```diff
@@ -0,0 +1,92 @@
...
+      def merge(timeout: nil, metadata: nil, retry_policy: nil)
+        @timeout ||= timeout
+        @metadata = metadata.merge(@metadata) if metadata
+        @retry_policy.merge(retry_policy) if @retry_policy.respond_to?(:merge)
+      end
+
+      ##
+      # @private
+      def set_default_values_for_internal_use!
+        @timeout ||= default_timeout
```
jbolinger: Where is `default_timeout` defined?

test/google/gax/operation_test.rb
```diff
@@ -28,12 +28,9 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 require 'test_helper'
-require 'time'
-
+# require 'time'
```
jbolinger: Is this intentional or meant to be removed?

lib/google/gax/api_call/retry_policy.rb
```diff
@@ -0,0 +1,113 @@
...
+        def initialize(retry_codes: nil, initial_delay: nil, multiplier: nil,
+                       max_delay: nil)
+          @retry_codes   = retry_codes
+          @initial_delay = initial_delay
+          @multiplier    = multiplier
+          @max_delay     = max_delay
+        end
+
+        def retry_codes
+          @retry_codes || [GRPC::Core::StatusCodes::UNAVAILABLE]
```
jbolinger: I think the default value for this should be empty. It's a bit counter-intuitive but `UNAVAILABLE` is not safe to retry on without additional information. It's possible for the server to have done some processing but still return an `UNAVAILABLE` status.

test/google/gax/api_call/retry_raise_test.rb
```diff
@@ -0,0 +1,136 @@
...
+
+    api_meth_stub = proc do |deadline: nil, **_kwargs|
+      deadline_arg = deadline
+      call_count += 1
+      raise GRPC::BadStatus.new(FAKE_STATUS_CODE_1, 'unknown')
+    end
+
+    api_call = Google::Gax::ApiCall.new(api_meth_stub)
+    options = Google::Gax::CallOptions.new(
+      retry_policy: { retry_codes: [14, 101] }
```
jbolinger: This is ok for tests, but it might help to put a note here if these codes are actual codes (i.e. what constant to they correspond to) or are out of range.

test/google/gax/api_call/retry_test.rb
```diff
@@ -0,0 +1,71 @@
...
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+require 'test_helper'
+
+class ApiCallRetryTest < Minitest::Test
```
jbolinger: Do we have any tests for custom retry policies? For example, a policy that never retries and maybe another that retries a fixed number of times?

lib/google/gax/api_call.rb
blowmage: You want the client method to calculate the header off the request object and update the CallOptions object before making the call?

blowmage: If so, can we open a new issue to make this change? I would prefer to do that in a separate PR as it seems orthogonal to retry.

lib/google/gax/call_options.rb
blowmage: Oops, this should have been removed. My original design changed greatly as the commit was implemented.

blowmage: Removed.

test/google/gax/operation_test.rb
blowmage: I was checking if it was needed. The only methods being used is `Time.now`, which does not need the require. The require adds `Time.parse`, which is not being used.

Removed.

test/google/gax/api_call/retry_test.rb
blowmage: Yes. Pushed a commit with the tests.

lib/google/gax/api_call.rb
jbolinger: More or less. Just saying that there is no reason for this proc to be so specific and I'd like to get the `x-goog-*` stuff out of here.

Separate change is fine - this is a minor thing.

lib/google/gax/api_call/retry_policy.rb
blowmage: I’ll make the default be empty. What about the other defaults? Are they okay?

jbolinger: Yeah, I think the others are fine.

lib/google/gax/api_call.rb
blowmage: Fixed.

lib/google/gax/api_call.rb
blowmage: Moved to #173.

test/google/gax/api_call/retry_raise_test.rb
blowmage: These tests are ports of the original tests, which were all written using fake errors and fake `code` values. I am not a huge fan of them either.

Do you approve of the overall approach? I don't want to spend time writing better tests if the approach is flawed.

lib/google/gax/api_call/retry_policy.rb
blowmage: Removed UNAVAILABLE.

test/google/gax/api_call/retry_raise_test.rb
jbolinger: Ah, ok. Overall approach is great - thanks for doing it! 

I do think we should address any omissions in the tests now though.

blowmage: Okay, tests have been updated to use real GRPC error codes, and to only retry GRPC::BadStatus errors.

lib/google/gax/api_call.rb
```diff
@@ -0,0 +1,135 @@
...
+          res = op.execute
+          yield op if block
+          res
+        rescue StandardError => error
+          if retry?(deadline)
+            retry if options.retry_policy.call(error)
+          end
+
+          if error.is_a? GRPC::BadStatus
+            error = Google::Gax.from_error(error).new('RPC failed')
```
blowmage: One of the changes in this implementation is that wrapped errors match the underlying GRPC error, instead of being a RetryError. But I think it makes more sense for the GaxError to match the BadStatus instead of a RetryError.

test/google/gax/api_call/retry/raise_test.rb
```diff
@@ -0,0 +1,149 @@
...
+
+    options = Google::Gax::CallOptions.new(
+      retry_policy: { retry_codes: [GRPC::Core::StatusCodes::UNAVAILABLE] }
+    )
+    assert_raises Google::Gax::GaxError do
+      api_call.call Object.new, options: options
+    end
+  end
+
+  def test_no_retry_with_fake_grpc_error
```
jbolinger: It might be useful to add a similar test here that adds a custom policy that does retry this error.

blowmage: Added.

PR #176, opened 2019-04-16
Add StreamInput

blowmage:
This class is very useful for controlling the lifecycle of streaming RPC calls.

### DIFF
--- a/lib/google/gax/stream_input.rb
+++ b/lib/google/gax/stream_input.rb
@@ -30,11 +30,11 @@
 module Google
   module Gax
     ##
-    # Adds requests to a stream and holds the stream open until {#close} is
-    # called.
+    # Manages requests for an input stream and holds the stream open until {#close} is called.
+    #
     class StreamInput
       ##
-      # dfdf
+      # Create a new input stream object to manage streaming requests and hold the stream open until {#close} is called.
       #
       # @param requests [Object]
       #
--- a/lib/google/gax/stream_input.rb
+++ b/lib/google/gax/stream_input.rb
@@ -78,7 +78,6 @@ module Google
         loop do
           request = @queue.pop
           break if request.equal? self
-          raise request if request.is_a? Exception
           yield request
         end
       end
--- a/lib/google/gax/stream_input.rb
+++ b/lib/google/gax/stream_input.rb
@@ -50,8 +50,12 @@ module Google
       #
       # @param request [Object]
       #
+      # @return [StreamInput] Returns self.
+      #
       def push request
         @queue.push request
+
+        self
       end
       alias << push
       alias append push
@@ -59,9 +63,12 @@ module Google
       ##
       # Closes the stream.
       #
+      # @return [StreamInput] Returns self.
+      #
       def close
         @queue.push self
-        nil
+
+        self
       end
 
       ##
--- a/lib/google/gax/stream_input.rb
+++ b/lib/google/gax/stream_input.rb
@@ -1,4 +1,4 @@
-# Copyright 2016, Google Inc.
+# Copyright 2019, Google Inc.
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


lib/google/gax/stream_input.rb
```diff
@@ -0,0 +1,87 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+module Google
+  module Gax
+    ##
+    # Adds requests to a stream and holds the stream open until {#close} is
+    # called.
+    class StreamInput
+      ##
+      # dfdf
```
jbolinger: ?

lib/google/gax/stream_input.rb
```diff
@@ -0,0 +1,87 @@
...
+      #
+      # @param requests [Object]
```
jbolinger: Add docs (comment below is probably enough).

lib/google/gax/stream_input.rb
```diff
@@ -0,0 +1,87 @@
...
+      # @yieldparam [Object] request The request object.
+      #
+      # @return [Enumerator] An Enumerator is returned if no block is given.
+      #
+      def to_enum
+        return enum_for :to_enum unless block_given?
+        loop do
+          request = @queue.pop
+          break if request.equal? self
+          raise request if request.is_a? Exception
```
jbolinger: Do we have a test for this?

lib/google/gax/stream_input.rb
```diff
@@ -0,0 +1,87 @@
...
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+module Google
+  module Gax
+    ##
+    # Adds requests to a stream and holds the stream open until {#close} is
```
jbolinger: Can we add (a little) more detail about why this class is needed and how it should be used with gRPC streams?

lib/google/gax/stream_input.rb
blowmage: It might be better for this line to be removed. In Google Cloud we used these objects for more than just streaming, and that is where this line comes from.

jbolinger: Removing would be fine.

Curious, what do you use this for?

lib/google/gax/stream_input.rb
blowmage: I agree. I would like @quartzmo's input on this as well. Perhaps we can pair on it today.

None of the existing methods have code examples showing usage. I haven't been prioritizing docs in general and code examples in particular because the details keep changing. But I am expecting to do an exhaustive docs pass once things are mote stable.

You can see this class being used in the [showcase echo tests](https://github.com/googleapis/gapic-generator-ruby/blob/400c6977f87486b082d7c9f0c72efe791d52b466/shared/test/showcase/echo_test.rb) on googleapis/gapic-generator-ruby#94.

lib/google/gax/stream_input.rb
blowmage: Removed.

lib/google/gax/stream_input.rb
blowmage: Updated.

lib/google/gax/stream_input.rb
blowmage: Updated.

lib/google/gax/stream_input.rb
blowmage: Created #177 to track documentation overhaul.

lib/google/gax/stream_input.rb
blowmage: We use this in tests, to mimic the behavior of when a streaming RPC method has an error. But that isn't applicable to StreamInput.

PR #178, opened 2019-04-16
Refactor ApiCall

### LINKED ISSUE #151: Update Longrunning::OperationsClient
Update Longrunning::OperationsClient to use the new generator and GAX code.

### LINKED ISSUE #165: More visibility/control in error retry

### LINKED ISSUE #173: Remove params_extractor
This functionality should be moved out of GAX to the client.

blowmage:
This PR does the following:

* Simplifies `ApiCall`
  * Remove `params_extractor` and `exception_transformer`.
  * Rename `ApiCall::Options` (was `CallOptions`).
    * Renames `Options#apply_defaults`
* Renames `RetryPolicy#apply_defaults`
  * Make this be more consistent with `Options#apply_defaults`.
* Adds the following procs to `ApiCall#call`:
  * `format_response` to format response objects.
    * Will be used to format longrunning and paged response objects.
  * `operation_callback` to yield operation objects
    * Useful for accessing trailing metadata.
  * `stream_callback` to yield every streamed response object.
* Update `Operations`
  * Add `ApiCall::Options` argument to RPC to methods.
  * Add Operation `RetryPolicy` and update `wait_for_done?` to use it.
* Update `OperationsClient`
* Remove `BackoffSettings`.
* Remove `MILLIS_PER_SECOND`.

[closes #151, closes #165, closes #173]


lib/google/gax/api_call.rb
```diff
@@ -45,74 +46,63 @@ class ApiCall
...
-          yield op if block
-          res
+          operation = stub_method.call request, deadline: deadline, metadata: options.metadata, return_op: true, &block
+
+          if on_stream
+            Thread.new { operation.execute }
+          else
+            response = operation.execute
+            response = on_response.call response if on_response
+            on_operation&.call response, operation
```
jbolinger: Do we need both `on_response` and `on_operation` (i.e. vs just ignoring the parameter)?

lib/google/gax/api_call.rb
```diff
@@ -45,74 +46,63 @@ class ApiCall
...
       ##
       # Invoke the API call.
       #
       # @param request [Object] The request object.
-      # @param options [CallOption, Hash] The options for making the API call.
-      # @param block [Proc] The proc to call when the API call is made.
+      # @param options [ApiCall::Options, Hash] The options for making the API call. A Hash can be provided to customize
+      #   the options object, using keys that match the arguments for {ApiCall::Options.new}. This object should only be
+      #   used once.
+      # @param on_response [Proc] A Proc object to format the response object. The Proc should accept response as an
```
jbolinger: Could do this during the doc overhaul, but a little more guidance on how each of these `on_*` blocks are used, if they are mutually exclusive or not, etc. would help.

lib/google/gax/api_call.rb
```diff
@@ -45,74 +46,63 @@ class ApiCall
...
+          end
         rescue StandardError => error
           if check_retry? deadline
             retry if options.retry_policy.call error
           end
 
           error = Google::Gax.from_error(error).new "RPC failed" if error.is_a? GRPC::BadStatus
 
-          raise error if @exception_transformer.nil?
-          @exception_transformer.call error
```
jbolinger: Can we also update https://github.com/googleapis/gapic-generator-ruby/pull/94 with the replacement for this?

lib/google/gax/api_call.rb
```diff
@@ -45,74 +46,63 @@ class ApiCall
...
         begin
-          op = @stub_method.call request, deadline: deadline, metadata: options.metadata, return_op: true
-
-          res = op.execute
-          yield op if block
-          res
+          operation = stub_method.call request, deadline: deadline, metadata: options.metadata, return_op: true, &block
+
+          if on_stream
+            Thread.new { operation.execute }
```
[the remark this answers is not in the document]
blowmage: It doesn't. Currently streaming requests are not retried and this matches existing behavior.

By returning the thread users can observe if/when the stream has an error or when it closes. I think this is the right approach, as if gives users the most control.

But, if we can think of a way to observe the thread and respond to errors in the GAX layer we can add it. But for things like Pub/Sub we need to know when the stream errors because we have to create the new stream based off the old stream's unsent contents. It's more complicated than anything Gax can do for us.

lib/google/gax/api_call.rb
blowmage: Yes, they perform two different behaviors, and are instantiated by different actions. `on_response` formats the response, and is used for LRO and Paging, and `on_operation` is for peeking at trailing metadata.

If a client method needs to wrap the response, it will send a `on_response` proc. If a user wants to peek at trailing metadata they pass a block to the client, which will be sent as the `on_operation` proc.

lib/google/gax/api_call.rb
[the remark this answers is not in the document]
jbolinger: I agree. I don't think we should do anything special here (other than document that they don't interact).

lib/google/gax/api_call.rb
blowmage: Agreed. Pairing with @quartzmo on the docs will be very beneficial, as he is great at pulling info out and communicating it clearly.

Usage is demonstrated in the tests.

lib/google/gax/api_call.rb
blowmage: We don't have a need for it in that PR, as we are only working on the Cloud generator in it. But, the feature is super easy to add. Assume you have a client method like the following:

```ruby
def echo request = nil, options: nil, **request_fields, &block
  # Lots of stuff commented out here...

  @echo ||= Google::Gax::ApiCall.new @echo_stub.method :echo
  @echo.call request, options: options, on_operation: block
end
```

All you need to add is a `rescue` statement, and wrap the raised error:

```ruby
def echo request = nil, options: nil, **request_fields, &block
  # Lots of stuff commented out here...

  @echo ||= Google::Gax::ApiCall.new @echo_stub.method :echo
  @echo.call request, options: options, on_operation: block
rescue Google::Gax::GaxError => error
  raise GoogleAdsError.new error.message
end
```

I really don't see a benefit for swallowing errors in the the Gax layer when the functionality is so easily added using a custom generator (which they are going to be doing anyway).

blowmage: To rephrase: this looks like an Ads-specific feature, and I don't see where it fits into the Cloud generator.

jbolinger: Ok, I buy this now. This kind of customization would have been more difficult in the past without the framework but I see your point - now we probably don't need a special knob for something like this.

lib/google/gax/api_call.rb
```diff
@@ -45,74 +46,63 @@ class ApiCall
...
-      end
-
-      def apply_params_extractor! request, options
-        return if @params_extractor.nil?
-
-        routing_header = calculate_routing_header request, @params_extractor
-        options.metadata["x-goog-request-params"] = routing_header
-      end
+      def compose_stream_proc on_stream: nil, on_response: nil
+        return unless on_stream
```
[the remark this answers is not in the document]
blowmage: No, this block is only passed to GRPC for streaming calls. The `on_response` will be applied for non-streaming responses by calling `#call` on it.

lib/google/gax/api_call.rb
```diff
@@ -45,74 +46,63 @@ class ApiCall
...
+        return on_stream unless on_response
 
-      def calculate_routing_header request, params_extractor
-        params = params_extractor.call request
-        params.map { |k, v| "#{k}=#{v}" }.join("&")
+        proc { |response| on_stream.call on_response.call response }
```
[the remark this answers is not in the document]
blowmage: AFAIK, we don't have any API that would use this today, but it is technically possible that this is something that might happen. Like, if the RPC was streaming longrunning operations responses that we wanted to format as GAX operations. Or, a user is using GAX outside of the GAPIC generator and wants to apply some wrapper to streamed responses.

lib/google/gax/api_call.rb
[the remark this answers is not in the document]
jbolinger: gotcha

lib/google/gax/api_call.rb
jbolinger: Ah, ok. So I was more confused than I thought...

Looking through the whole method again, and ignoring the names, what you're doing seems fine. The naming I find quite counter-intuitive though, particularly `response` / `on_response` (which can mean either the raw or formatted response) and `on_operation` (which sounds like LRO but it's not).

I am not great at names, but something like this would be clearer to me:
`on_response` -> `format_response` or `on_raw_response` or `on_grpc_response` or `on_data`, etc.
`on_operation` -> `on_complete` or `on_done`, `on_result`, etc.
`on_stream` -> this one is clear as-is. Perhaps `on_next_result` or similar for consistency depending on the other names used.

From a lifecycle perspective, I like this:
`on_data` -> `on_done` (unary)
`on_data` -> `on_stream` -> `on_data` -> `on_stream` -> `...` ->`on_done` (streaming)
rather than this, which sounds backwards to me:
`on_response` -> `on_operation`

blowmage: What do you think about the names `format_response`, `operation_callback`, and `stream_callback`?

lib/google/gax/api_call.rb
blowmage: I have added some code examples for now. PTAL.

lib/google/gax/api_call.rb
jbolinger: I like that and I think it's much clearer. Is `*_callback` a more common convention than `on_*`? I have no preference and leave it up to you - just curious.

It's probably worth mentioning in the doc comments that the `operation` is a grpc operation (maybe reference the type?) so it's not confused with the LRO concept that we also have floating around in this lib.

blowmage:
> this kind of simplification is long overdue!

There is much more coming...

lib/google/longrunning/operations_client.rb
```diff
@@ -172,14 +136,26 @@ def initialize \
...
-        req = Google::Gax.to_proto req, Google::Longrunning::GetOperationRequest
-        @get_operation.call req, options
+      def get_operation request = nil, options: nil, **request_fields, &block
+        raise ArgumentError, "request must be provided" if request.nil? && request_fields.empty?
+        if !request.nil? && !request_fields.empty?
+          raise ArgumentError, "cannot pass both request object and named arguments"
+        end
+
+        request ||= request_fields
+        # request = Google::Gax::Protobuf.coerce request, to: Google::Showcase::V1alpha3::EchoRequest
```
jbolinger: remove?

lib/google/longrunning/operations_client.rb
```diff
@@ -172,14 +136,26 @@ def initialize \
...
+        request = Google::Gax.to_proto request, Google::Longrunning::GetOperationRequest
+
+        # Converts hash and nil to an options object
+        options = Google::Gax::ApiCall::Options.new options.to_h if options.respond_to? :to_h
+        echo_header_params = { name: request.name }
```
jbolinger: I think you may have let some stuff from showcase leak into this one?

lib/google/longrunning/operations_client.rb
```diff
@@ -172,14 +136,26 @@ def initialize \
...
+        echo_request_params_header = echo_header_params.map { |k, v| "#{k}=#{v}" }.join("&")
+        echo_metadata = @metadata.merge "x-goog-request-params" => echo_request_params_header
+        echo_retry_policy = { retry_codes: [GRPC::Core::StatusCodes::UNAVAILABLE] }
```
jbolinger: This is probably safe to do in LRO methods, but it might be wise to open an issue to verify this (if we don't already have one - I know I bring this up a lot).

lib/google/longrunning/operations_client.rb
```diff
@@ -295,30 +291,67 @@ def cancel_operation \
...
+          interceptors: interceptors,
+          &stub_new
+        )
+      end
 
-        { "x-goog-api-client" => google_api_client }
+      def x_goog_api_client_header lib_name, lib_version
+        x_goog_api_client_header = ["gl-ruby/#{RUBY_VERSION}"]
+        x_goog_api_client_header << "#{lib_name}/#{lib_version}" if lib_name
+        x_goog_api_client_header << "gapic/#{Google::Showcase::VERSION}"
```
jbolinger: ?

lib/google/longrunning/operations_client.rb
```diff
@@ -295,30 +291,67 @@ def cancel_operation \
...
-        google_api_client << "#{lib_name}/#{lib_version}" if lib_name
-        google_api_client << "gapic/#{package_version}"
-        google_api_client << "gax/#{Google::Gax::VERSION}"
-        google_api_client << "grpc/#{GRPC::VERSION}"
-        google_api_client.join " "
+        # Allow overriding the service path/port in subclasses.
+        service_path = self.class::SERVICE_ADDRESS
+        port = self.class::DEFAULT_SERVICE_PORT
+        interceptors = self.class::GRPC_INTERCEPTORS
+        stub_new = Google::Showcase::V1alpha3::Echo::Stub.method :new
```
jbolinger: Echo?

blowmage:
The GAPIC side of this change can be found in googleapis/gapic-generator-ruby#108.

lib/google/longrunning/operations_client.rb
blowmage: Removed.

lib/google/longrunning/operations_client.rb
blowmage: Yeah, this was copied from the wrong local branch. Fixed now.

lib/google/longrunning/operations_client.rb
blowmage: Removed.

lib/google/longrunning/operations_client.rb
blowmage: Removed.

lib/google/longrunning/operations_client.rb
blowmage: Yeah, old version. Fixed.

PR #183, opened 2019-04-24
2.0/update errors

blowmage:
* Update GaxError to Ruby 2.3
  * Remove Ruby 2.0 specific code since minimum Ruby is now 2.3.
  * Move `GaxError#status_details` to `google/gax/grpc/status_details.rb` file, so that the method is only loaded when GRPC is.


lib/google/gax/grpc/status_details.rb
```diff
@@ -0,0 +1,60 @@
...
+# Required in order to deserialize common error detail proto types
+require "google/rpc/error_details_pb"
+
+module Google
+  module Gax
+    class GaxError < StandardError
+      def status_details
+        return unless cause.is_a? GRPC::BadStatus
+
+        rpc_status = GRPC::GoogleRpcStatusUtils.extract_google_rpc_status cause.to_status
```
blowmage: In the future, we will be able to replace this with just `cause.to_rpc_status`.

lib/google/gax/api_call.rb
```diff
@@ -158,7 +158,8 @@ def call request, options: nil, format_response: nil, operation_callback: nil, s
             retry if options.retry_policy.call error
           end
 
-          error = Google::Gax.from_error(error).new "RPC failed" if error.is_a? GRPC::BadStatus
+          # Wrap error in GaxError if GRPC::BadStatus
+          error = Google::Gax::GaxError.wrap error
```
jbolinger: nit: this reads less clear than the original to me. I prefer the check here and perhaps a raise if the error is inappropriate. (I'm pointing this out due to my other comment about keeping this public, otherwise it doesn't really matter.)

lib/google/gax/errors.rb
```diff
@@ -67,6 +56,22 @@ def metadata
         return nil unless cause&.respond_to? :metadata
         cause.metadata
       end
+
+      ##
+      # @private Wrap GRPC error with associated GAX error object.
```
jbolinger: I'm not sure we should exclude this from the public API. It might be useful for some kinds of customization and I don't think it adds much complexity to expose and document it. Any reason you want to keep it?

lib/google/gax/grpc/status_details.rb
```diff
@@ -0,0 +1,60 @@
...
+        rpc_status.details.map do |detail|
+          begin
+            detail_type = Google::Protobuf::DescriptorPool.generated_pool.lookup detail.type_name
+            detail.unpack detail_type.msgclass
+          rescue StandardError
+            detail
+          end
+        end
+      rescue StandardError
+        nil
```
jbolinger: Is this meant to replace the previous error message about malformed responses? Can we do any better than returning `nil` here?

blowmage: We can raise an error. I didn't understand why the previous implementation would return a string when malformed, instead of allowing the `Google::Protobuf::ParseError` to be raised. Having to check for a single string instead of an array is not significantly different than handling an error.

lib/google/gax/errors.rb
blowmage: I can't think of a use case for why a user would want to wrap a `GRPC::BadStatus` error with a `Google::Gax::GaxError`. It seems like internal logic to me.

lib/google/gax/api_call.rb
blowmage: This is the only place this method is used. Perhaps we should move the implementation from `Google::Gax::GaxError` to a private method on `Google::Gax::ApiCall`?

lib/google/gax/errors.rb
jbolinger: Interceptors was the primary use case that I had in mind, but agreed it will be rare.

IIRC Ads was also creating `GaxError` instances in order to extract the status details so we should probably keep that helper exposed.

lib/google/gax/grpc/status_details.rb
jbolinger: Yeah, the old behavior was confusing (we had a recently complain from Sam actually) so that sounds more reasonable to me.

lib/google/gax/grpc/status_details.rb
blowmage: `BadStatus#to_rpc_status` is being added in grpc/grpc#18752. It currently handles `Google::Protobuf::ParseError` errors and returns `nil`.

lib/google/gax/api_call.rb
jbolinger: Your call. That sounds better to me though.

lib/google/gax/errors.rb
blowmage: K, reverted.

lib/google/gax/api_call.rb
blowmage: Reverted.

blowmage: Added `wrap_error` as a private method.

lib/google/gax/grpc/status_details.rb
```diff
@@ -0,0 +1,60 @@
+# Copyright 2016, Google Inc.
```
dazuma: New files, update to Copyright 2019 Google LLC

lib/google/gax/grpc/status_details.rb
dazuma: I would prefer to raise an error, or in general not to drop the error on the floor.

There are other cases where `nil` is returned. Since the happy path for this method returns an array, can we have all code paths either return an array (even an empty one) or raise an error? It's weird to return a collection or nil; it means callers can't just call enumerable methods on the result like they might want to, but have to add nil checks.

lib/google/gax/grpc/status_details.rb
blowmage: Updated.

lib/google/gax/grpc/status_details.rb
quartzmo: @blowmage Thanks for explaining the intention to replace this in the future. Can you add a TODO comment here as well to that effect?

lib/google/gax/grpc/status_details.rb
```diff
@@ -0,0 +1,60 @@
...
+
+        rpc_status.details.map do |detail|
+          begin
+            detail_type = Google::Protobuf::DescriptorPool.generated_pool.lookup detail.type_name
+            detail.unpack detail_type.msgclass
+          rescue StandardError
+            detail
+          end
+        end
+      rescue StandardError
```
blowmage:
```suggestion
      rescue Google::Protobuf::ParseError
```

quartzmo: Can this error declaration be narrowed?

lib/google/gax/grpc/status_details.rb
```diff
@@ -0,0 +1,60 @@
...
+
+        rpc_status = GRPC::GoogleRpcStatusUtils.extract_google_rpc_status cause.to_status
+
+        return if rpc_status.nil?
+
+        rpc_status.details.map do |detail|
+          begin
+            detail_type = Google::Protobuf::DescriptorPool.generated_pool.lookup detail.type_name
+            detail.unpack detail_type.msgclass
+          rescue StandardError
```
quartzmo: Can this error declaration be narrowed?

lib/google/gax/grpc/status_details.rb
dazuma: I'm not clear where this can be raised. If we can narrow the type, I'd expect we can narrow the scope as well. Either that or it's meant to be a catch-all error handler to prevent exceptions from propagating...?

blowmage: It is meant to capture a `ParseError`, but the original code used `StandardError`. If you look at the original code the rescue is specific to the `GRPC::GoogleRpcStatusUtils.extract_google_rpc_status` call. We can remove this entirely if we update the minimum grpc dependency to a version that includes grpc/grpc#18752 (when released).

blowmage: Yes, I'll move it closer to the call it is rescuing from.

blowmage: Updated. PTAL.

blowmage: Updated. PTAL.

lib/google/gax/grpc/status_details.rb
blowmage: Added. PTAL.

lib/google/gax/grpc/status_details.rb
blowmage: Updated. PTAL.

lib/google/gax/grpc/status_details.rb
blowmage: We discussed raising or not raising in grpc/grpc#18752. My view here is that this is a convenience method, and it is more convenient to not have to add error handling when trying to pull values from the object.

FWIW, I do agree that it would be nicer to return an empty array here instead of nil. Should I make that change?

blowmage: The justification for returning `nil` here is that it matches the behavior of the other error methods.

```ruby
      def code
        return nil unless cause&.respond_to? :code
        cause.code
      end

      def details
        return nil unless cause&.respond_to? :details
        cause.details
      end

      def metadata
        return nil unless cause&.respond_to? :metadata
        cause.metadata
      end
```

dazuma: Yeah I think I'd prefer to see empty arrays/hashes for all those. (i.e. if the cause conceptually has no status details, then return an empty array meaning no details.) For `code` I'm not sure. If that's numeric, then maybe `nil` is reasonable to indicate absence of value.

dazuma: The rest of this PR LGTM. I think I'm just stuck on the return values for the convenience methods. Especially if they're intended as convenience methods they should be... convenient.

blowmage: @dazuma Updated. PTAL.

PR #184, opened 2019-05-02
Refactor Proto.coerce

blowmage:
This used to be a class method `Gax.to_proto`, but the name had issues. First, it was not a conversion method on an object, but on a class. The `to_*` methods are usually used to convert from one object to another object. Renaming to `coerce` is more accurate.


lib/google/gax/protobuf.rb
```diff
@@ -0,0 +1,169 @@
+# Copyright 2019, Google Inc.
```
dazuma: `Google LLC` please

blowmage: Should I change the company attribution on all files? Should we also change the 1.x branch?

blowmage: I'm going to spin this out into its own issue. #187

PR #189, opened 2019-05-04
Operation Updates

blowmage:
Use the new `Operations` class from gapic generator. Update the constructor to use named arguments for optional arguments (`result_type` and `metadata_type`).

### DIFF
--- a/lib/google/gax/operation.rb
+++ b/lib/google/gax/operation.rb
@@ -214,7 +214,7 @@ module Google
         # Converts hash and nil to an options object
         options = ApiCall::Options.new options.to_h if options.respond_to? :to_h
 
-        @client.cancel_operation @grpc_op.name, options: options
+        @client.cancel_operation name: @grpc_op.name, options: options
       end
 
       ##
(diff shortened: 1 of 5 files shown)


blowmage:
This PR cannot be merged until #190 is merged.

lib/google/gax/operation.rb
```diff
@@ -214,7 +214,7 @@ def cancel options: nil
         # Converts hash and nil to an options object
         options = ApiCall::Options.new options.to_h if options.respond_to? :to_h
 
-        @client.cancel_operation @grpc_op.name, options: options
+        @client.cancel_operation name: @grpc_op.name, options: options
```
dazuma: These were changed on the generated code side, I take it? It seems `name` would be a required parameter; why did we change this?

blowmage: This is part of the changes in the new generator. Client methods are not backwards compatible, as all field arguments are

dazuma: Oh, that's right, I forgot, these are client methods. Okay then, carry on. 👍

blowmage:
The google-style update is messing with the build. I'm gonna fix it in this PR.

PR #190, opened 2019-05-04
Add PagedEnumerable format_resource proc

blowmage:


lib/google/gax/paged_enumerable.rb
```diff
@@ -67,12 +67,15 @@ class PagedEnumerable
       # @param request_page_token_field [String] The name of the field in request which will have the page token.
       # @param response_page_token_field [String] The name of the field in the response which holds the next page token.
       # @param resource_field [String] The name of the field in the response which holds the resources.
+      # @param format_resource [Proc] A Proc object to format the resource object. The Proc should accept response as an
+      #   argument, and return a formatted resource object. Optional.
```
quartzmo: It should return a string, not an "object", correct? Maybe "return the correctly formatted type name [String] of the resource."

quartzmo: Might also be good: "For example: "Google::Cloud::PubSub::Topic""

dazuma: I assume this could "format" as objects other than strings, right? I think the initial use case here is for wrapping the resource in a `Google::Gax::Operation` object (see https://github.com/googleapis/gapic-generator-ruby/pull/136)

blowmage: It should not return a string. The only current use is to format a `Google::Longrunning::Operation` resource object as a `Google::Gax::Operation` object. The same way the `format_response` proc formats a `Google::Longrunning::Operation` response object as a `Google::Gax::Operation` object for a `get_operation` RPC call. The difference here is that the formatting is being applied to each resource the `PagedEnumerable` is enumerating.

quartzmo: Sounds OK for now.

blowmage: That said, I would not be opposed to a better term that “format”.

PR #194, opened 2019-05-10
Add Configuration

blowmage:
This PR adds `Google::Gax::Configuration` which will be used to configure the Client defaults. This will allow users to configure the behavior of API client, by both configuring the defaults on the global level as well as configuring the defaults on the client instance object level.

```ruby
# Set global defaults
Google::Cloud::Vision::V1::ImageAnnotator.configure do |config|
  config.timeout = 120
  config.retry_policy = { retry_codes: [14] }
  config.methods.batch_annotate_images.retry_policy = { retry_codes: [12, 14] }
end

# Set/override the defaults on an individual instance
client = Google::Cloud::Vision::V1::ImageAnnotator::Client.new do |config|
  config.timeout = 90
  config.retry_policy = { retry_codes: [12, 14] }
  config.methods.batch_annotate_images.retry_policy = { retry_codes: [14] }
end
```

[refs #188]

### DIFF
--- a/lib/google/gax/configuration.rb
+++ b/lib/google/gax/configuration.rb
@@ -44,13 +44,12 @@ module Google
     #
     # Note that config objects inherit from `BasicObject`. This means it does
     # not define many methods you might expect to find in most Ruby objects.
-    # For example, `to_s`, `inspect`, `is_a?`, `instance_variable_get`, and so
-    # forth.
+    # For example, `to_s`, `is_a?`, `instance_variable_get`, and so forth.
     #
     # @example
     #   require "google/gax/configuration"
     #
-    #   config = Google::Gax::Configuration.create do |c|
+    #   config = Google::Gax::Configuration.new do |c|
     #     c.add_field! :opt1, 10
     #     c.add_field! :opt2, :one, enum: [:one, :two, :three]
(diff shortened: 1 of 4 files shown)


quartzmo (approved):

lib/google/gax/configuration.rb
```diff
@@ -0,0 +1,593 @@
...
+        @defaults = {}
+        @validators = {}
+        add_options legacy_categories
+      end
+
+      ##
+      # Legacy method of adding subconfigs. This is used by older versions of
+      # the stackdriver client libraries but should not be used in new code.
+      #
+      # @deprecated
```
dazuma: Can we just remove these deprecated/legacy methods since we don't have existing users of this copy of the class?

(If we remove, also remove `:add_options` from `ILLEGAL_KEYS` below.)

dazuma (approved):
Heh...

lib/google/gax/configuration.rb
blowmage: LOL, I just copied/pasted because that's who I am.

dazuma: "Good developers copy. Great developers paste." —Kelsey Hightower 🤣

PR #195, opened 2019-05-11
 Add DerivedConfiguration

blowmage:
This new class shares the structure with a super configuration object,
but it can only change values. It cannot alter the structure.

The intention is that this will be used to customize a global config.
For example, to override settings for an individual client.


lib/google/gax/configuration.rb
```diff
@@ -112,15 +116,21 @@ def self.config? obj
...
 
         # Can't call yield because of BasicObject
         block&.call self
       end
 
+      ##
+      # Derive a Configuration object. The subsequent object can set local values,
+      # but will not be able to change the structure.
+      #
+      def derive! &block
```
blowmage: I do not like this name. I really don't. But I can't think of anything better and so I went with it. Hopefully someone can suggest a far superior name for what this does.

lib/google/gax/configuration/derived_configuration.rb
```diff
@@ -0,0 +1,157 @@
...
+      # is attempted on a derived configuration
+      #
+      class DerivedConfigurationError < ::StandardError
+      end
+
+      ##
+      # A Derived Configuration object. It can set local values,
+      # but cannot change the configuration structure.
+      #
+      class DerivedConfiguration < Configuration
```
blowmage: Thinking on it over the weekend, I think I might prefer `Google::Gax::Configuration::Derived` to `Google::Gax::Configuration::DerivedConfiguration`.

lib/google/gax/configuration/deferred_value.rb
```diff
@@ -0,0 +1,60 @@
...
+    class Configuration < BasicObject
+      ##
+      # Create a configuration value that will be invoked when retrieved.
+      #
+      def self.deferred &block
+        DeferredValue.new(&block)
+      end
+
+      ##
+      # A configuration value that is be invoked as it is retrieved.
```
blowmage: I flipped this to public because it seems like an important behavior and something that should be documented.

lib/google/gax/configuration/schema.rb
```diff
@@ -0,0 +1,338 @@
...
+        #     validation fails. Defaults to `true`.
+        #
+        def initialize show_warnings: true
+          @show_warnings = show_warnings
+          @defaults = {}
+          @validators = {}
+        end
+
+        ##
+        # Add a value field to this configuration.
```
dazuma: This is a private class. I don't think we should copy all the documentation details; they're just going to get stale.

test/google/gax/configuration_test.rb
```diff
@@ -88,7 +88,11 @@
     end
 
     it "can inspect" do
-      assert_equal "#<Google::Gax::Configuration {:k1=>{:k2=>{:k3=>{}}}}>", simple_config.inspect
+      expected_inspect = "#<Google::Gax::Configuration: k1=" \
+        "<Google::Gax::Configuration: k2=" \
+        "<Google::Gax::Configuration: k3=" \
+        "<Google::Gax::Configuration: >>>>"
```
dazuma: This seems less useful than the hash-like representation

lib/google/gax/configuration/derived_configuration.rb
```diff
@@ -0,0 +1,157 @@
...
+        # Constructs a Configuration object. If a block is given, yields `self` to the
+        # block, which makes it convenient to initialize the structure by making
+        # calls to `add_field!` and `add_config!`.
+        #
+        # @param [boolean] show_warnings Whether to print warnings when a
+        #     validation fails. Defaults to `true`.
+        #
+        def initialize super_config, &block
+          @values = {}
+          @config = super_config
```
dazuma: Maybe calling this `@super_config` or `@parent_config` would be more clear.

lib/google/gax/configuration/deferred_value.rb
dazuma: Isn't it an implementation detail, though? Unless we expect that users will generally use `DeferredValue.new` instead of `Configuration.deferred` (in which case maybe we should drop the latter)

lib/google/gax/configuration.rb
```diff
@@ -231,16 +235,16 @@ def reset! key = nil
         if key.nil?
           @values.each_key { |k| reset! k }
         else
-          key = key.to_sym
-          if @defaults.key? key
-            @values[key] = @defaults[key]
-            @values[key].reset! if @validators[key] == SUBCONFIG
+          key = ::Kernel.String(key).to_sym
```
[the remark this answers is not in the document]
blowmage: All keys are converted to symbols today. The current code uses `key.to_s.to_sym`, but I replaced it with `::Kernel.String(key).to_sym` so that an error would be raised if a key object was ever passed in that wasn't able to be converted to a string.

lib/google/gax/configuration/schema.rb
blowmage: Agreed.

test/google/gax/configuration_test.rb
blowmage: Agreed. I'll change it.

lib/google/gax/configuration/derived_configuration.rb
blowmage: Agreed. Renamed.

test/google/gax/configuration_test.rb
blowmage: Changed.

lib/google/gax/configuration/schema.rb
blowmage: Updated.

blowmage:
> Isn't it an implementation detail, though? Unless we expect that users will generally use DeferredValue.new instead of Configuration.deferred (in which case maybe we should drop the latter)

Both were marked private. If `Configuration.deferred` is marked visible then we also have to document it's return type, which is `DeferredValue`. I'd be fine marking `DeferredValue.new` as private, but previous to this both `Configuration.deferred` and `DeferredValue` were marked private.

lib/google/gax/configuration.rb
dazuma: I don't mind it. OO terminology seems appropriate.

lib/google/gax/configuration/derived_configuration.rb
dazuma: I agree.

dazuma:
> > Isn't it an implementation detail, though? Unless we expect that users will generally use DeferredValue.new instead of Configuration.deferred (in which case maybe we should drop the latter)
> 
> Both were marked private. If `Configuration.deferred` is marked visible then we also have to document it's return type, which is `DeferredValue`. I'd be fine marking `DeferredValue.new` as private, but previous to this both `Configuration.deferred` and `DeferredValue` were marked private.

Then I'm inclined to prefer keeping `DeferredValue.new`, and dropping `Configuration.deferred`. The latter is not as clear, and I'm not sure what benefit it gives us.

lib/google/gax/configuration/derived_configuration.rb
blowmage: Updated.

lib/google/gax/configuration/deferred_value.rb
blowmage: > Then I'm inclined to prefer keeping `DeferredValue.new`, and dropping `Configuration.deferred`. The latter is not as clear, and I'm not sure what benefit it gives us.

I agree. I removed `Configuration.deferred`.

lib/google/gax/configuration.rb
```diff
@@ -112,15 +116,21 @@ def self.config? obj
...
-        @validators = {}
+        @schema = Schema.new show_warnings: show_warnings
 
         # Can't call yield because of BasicObject
         block&.call self
       end
 
+      ##
+      # Derive a Configuration object. The subsequent object can set local values,
+      # but will not be able to change the structure.
```
quartzmo: Can you document the block and explain its usage? (An example would probably be helpful too.)

quartzmo: I think the best example would show how the block enables a "DSL"-type configuration style suitable for use in a Rails initializer, for example.

dazuma (approved):

lib/google/gax/configuration.rb
blowmage: I've updated the docs, but didn't add any code examples.

PR #196, opened 2019-05-17
Add Config DSL

blowmage:
This PR is an alternate implementation of the Configuration library. It provides a single module that clients can use to easily create a `Configuration` class. It is parent aware, so nested/child configuration objects can be created that will look to the parent object for values when local values are not present.

### DIFF
--- a/lib/google/gax/config.rb
+++ b/lib/google/gax/config.rb
@@ -83,9 +83,6 @@ module Google
           if instance_variable_defined? :@parent_config
             parent = instance_variable_get :@parent_config
             return parent.send name if parent&.respond_to? name
-
-            parent = instance_variable_get :@parent_config
-            return parent.send name if parent&.respond_to? name
           end
 
           default
--- a/lib/google/gax/config.rb
+++ b/lib/google/gax/config.rb
@@ -38,67 +38,65 @@ module Google
       # @param [Hash] opts Validation options
       #
       def config_attr name, default, *valid_values, &validator
-         name = String(name).to_sym
-         raise "cannot create field named parent_config" if name == :parent_config
-         raise "method #{name} already exists" if method_defined? name
+        name = String(name).to_sym
+        raise "cannot create field named parent_config" if name == :parent_config
+        raise "method #{name} already exists" if method_defined? name
 
-         raise "validation must be provided" if validator.nil? && valid_values.empty?
-         validator ||= ->(value) { valid_values.any? { |v| v === value } }
+        raise "validation must be provided" if validator.nil? && valid_values.empty?
+        validator ||= ->(value) { valid_values.any? { |v| v === value } }
 
-         name_ivar = "@#{name}".to_sym
+        name_ivar = "@#{name}".to_sym
 
-         if default.nil?
-           # create getter
-           define_method name do
-             value = instance_variable_get name_ivar
-             if value.nil?
-               parent = instance_variable_get :@parent_config
-               if parent&.respond_to? name
-                 return parent.send name
-               end
-             end
-             value
-           end
-           # create setter
-           define_method "#{name}=" do |new_value|
-             parent = instance_variable_get :@parent_config
-             valid_value = validator.call new_value
-             # Allow nil if parent config has the same method.
-             valid_value = true if new_value.nil? && parent&.respond_to?(name)
-             raise ArgumentError unless valid_value
-             if new_value.nil?
-               remove_instance_variable name_ivar if instance_variable_defined? name_ivar
-             else
-               instance_variable_set name_ivar, new_value
-             end
-           end
-         else
-           # create getter with default value
-           define_method name do
-             value = instance_variable_get name_ivar
-             if value.nil?
-               parent = instance_variable_get :@parent_config
-               if parent&.respond_to? name
-                 return parent.send name
-               else
-                 return default
-               end
-             end
-             value
-           end
-           # create setter with default value
-           define_method "#{name}=" do |new_value|
-             valid_value = validator.call new_value
-             # Always allow nil because we have a default value
-             valid_value = true if new_value.nil?
-             raise ArgumentError unless valid_value
-             if new_value.nil?
-               remove_instance_variable name_ivar if instance_variable_defined? name_ivar
-             else
-               instance_variable_set name_ivar, new_value
-             end
-           end
-         end
+        if default.nil?
+          # create getter
+          define_method name do
+            value = instance_variable_get name_ivar
+            if value.nil?
+              parent = instance_variable_get :@parent_config
+              return parent.send name if parent&.respond_to? name
+            end
+            value
+          end
+          # create setter
+          define_method "#{name}=" do |new_value|
+            parent = instance_variable_get :@parent_config
+            valid_value = validator.call new_value
+            # Allow nil if parent config has the same method.
+            valid_value = true if new_value.nil? && parent&.respond_to?(name)
+            raise ArgumentError unless valid_value
+            if new_value.nil?
+              remove_instance_variable name_ivar if instance_variable_defined? name_ivar
+            else
+              instance_variable_set name_ivar, new_value
+            end
+          end
+        else
+          # create getter with default value
+          define_method name do
+            value = instance_variable_get name_ivar
+            if value.nil?
+              parent = instance_variable_get :@parent_config
+              if parent&.respond_to? name
+                return parent.send name
+              else
+                return default
+              end
+            end
+            value
+          end
+          # create setter with default value
+          define_method "#{name}=" do |new_value|
+            valid_value = validator.call new_value
+            # Always allow nil because we have a default value
+            valid_value = true if new_value.nil?
+            raise ArgumentError unless valid_value
+            if new_value.nil?
+              remove_instance_variable name_ivar if instance_variable_defined? name_ivar
+            else
+              instance_variable_set name_ivar, new_value
+            end
+          end
+        end
       end
     end
   end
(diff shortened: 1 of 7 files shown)


lib/google/gax/config.rb
```diff
@@ -0,0 +1,127 @@
...
+            end
+          end
+        else
+          # create getter with default value
+          define_method name do
+            value = instance_variable_get name_ivar
+            if value.nil?
+              parent = instance_variable_get :@parent_config
+              return parent.send name if parent&.respond_to? name
+              return default
```
dazuma: This line seems to be the only difference between this case and the `default.nil?` case above. But they have the same behavior. i.e. if `default` and `value` are both nil, then nil will be returned here in this version of the method -- and in the `default.nil?` method above, nil will be returned at the end when `value` is returned. So I don't think we need two versions of the method. This bottom form will work in both cases.

lib/google/gax/config.rb
```diff
@@ -0,0 +1,127 @@
...
+    module Config
+      ##
+      # Add configuration attribute methods to the configuratin class.
+      #
+      # @param [String, Symbol] key The name of the option
+      # @param [Object, nil] initial Initial value (nil is allowed)
+      # @param [Hash] opts Validation options
+      #
+      def config_attr name, default, *valid_values, &validator
+        name = String(name).to_sym
```
dazuma: We should validate the name, and make sure it's a valid method name.

lib/google/gax/config.rb
```diff
@@ -0,0 +1,127 @@
...
+
+        raise "validation must be provided" if validator.nil? && valid_values.empty?
+        validator ||= ->(value) { valid_values.any? { |v| v === value } }
+
+        name_ivar = "@#{name}".to_sym
+
+        if default.nil?
+          # create getter
+          define_method name do
+            value = instance_variable_get name_ivar
```
dazuma: I think this will still generate warnings until someone sets the instance variable. Is there a way to make sure the initialize method initializes all instance variables (maybe including `@parent_config`) automatically?

lib/google/gax/config.rb
```diff
@@ -0,0 +1,127 @@
...
+    #
+    module Config
+      ##
+      # Add configuration attribute methods to the configuratin class.
+      #
+      # @param [String, Symbol] key The name of the option
+      # @param [Object, nil] initial Initial value (nil is allowed)
+      # @param [Hash] opts Validation options
+      #
+      def config_attr name, default, *valid_values, &validator
```
dazuma: After looking at how default and parent interact below, I wonder if there's a way can do without default altogether, and just rely on specifying/setting a parent. Just a thought. I think it may make things conceptually simpler, but it may be more complicated to implement and still have a reasonable usage experience.

dazuma (commented):
I like that the approach is a lot simpler. It assumes that the "schema" is static, but in our case I think it is. I do have some questions though.

lib/google/gax/config.rb
blowmage: I agree this is a valid refactoring that can be made with the code as it exists now.

lib/google/gax/config.rb
```diff
@@ -0,0 +1,127 @@
...
+        raise "validation must be provided" if validator.nil? && valid_values.empty?
+        validator ||= ->(value) { valid_values.any? { |v| v === value } }
+
+        name_ivar = "@#{name}".to_sym
+
+        if default.nil?
+          # create getter
+          define_method name do
+            value = instance_variable_get name_ivar
+            if value.nil?
```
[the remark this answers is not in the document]
blowmage: Correct. This is a tradeoff I made trying to simplify.

lib/google/gax/config.rb
```diff
@@ -0,0 +1,127 @@
...
+              parent = instance_variable_get :@parent_config
```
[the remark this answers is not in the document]
blowmage: We may not need to use `instance_variable_get`/`instance_variable_set`. These calls were first used from an earlier implementation trying use `instance_variables` like the `@values` has in Configuration. But eventually I found `method_defined?` was a much simpler way of checking.

I decided to keep those method calls because this change also uses `remove_instance_variable` to indicate that the parent config should be checked instead of the instance variable. Just in case we decided we wanted to set a `nil` value on a local/derived config object. This would be analogous to `@values.delete key`.

lib/google/gax/config.rb
blowmage: The `config_attr` call will raise a `NameError` if the name is not valid. This is something that `define_method` does.

lib/google/gax/config.rb
blowmage: Dang it, I just checked and you are correct. I thought this was not raising a warning. I'll refactor and add a `instance_variable_defined?` conditional.

lib/google/gax/config.rb
blowmage: I was kinda wondering the same. But having a default is super useful if we ever want to reset to it. e.g. a user changes the service host to something else, but wants to reset to the original value. If the config object doesn't maintain the default then I think we still want to set a default in a ruby constant. e.g. `Client::SERVICE_ADDRESS = "https://service.googleapis.com"`

lib/google/gax/config.rb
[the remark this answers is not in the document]
blowmage: Alternatively, we could use a different magic value to indicate setting the config value to the "pass through to parent" when a parent config us present, or "reset to default" otherwise.

```ruby
client.configure.host = :DEFAULT
```

lib/google/gax/config.rb
blowmage: This has been refactored and the two separate method definitions have been combined.

lib/google/gax/config.rb
blowmage: The warnings have been resolved.

lib/google/gax/config.rb
dazuma: Yes, I agree we do want a "static" default, but the question is, where does it live? For example, the logic I'd expect to see is:

* Value set in the current config? Then return it.
* Value set in the parent config? Then return it. (And continue up the ancestry.)
* Got to the toplevel config and still nothing? Then return the default defined statically somewhere.

Currently that final default is declared in the `config_attr` declaration. But that brings up the question of *whose* `config_attr` declaration? It is currently possible for a config's parent to have a different class, which could have the same field names with different defaults. The current logic (as best as I can read it) says, start with the toplevel ancestor, and work backwards towards the most detailed override, and take the first non-nil default. (i.e. defaults are chosen in reverse order from values.) I don't think this is intuitive behavior. We could solve this by enforcing that a parent config must have the same class (and thus the same fields and defaults) as a child config. But I'm not sure if you intended to allow duck-typing in the parent.

lib/google/gax/config.rb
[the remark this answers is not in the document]
dazuma: `nil` is probably as good as anything for this purpose. If we were to choose a different path, I think the alternative is to have an explicit separate `unset_<field-name>` method or something like it, and I think we both prefer the simplicity of `nil`.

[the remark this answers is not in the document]
blowmage: Right. Another idea I had was something like `config.reset! :name` or `config.default! :name`, but I think that approach adds more friction than using `nil` the way it is being used now. But I recognize that my comfort level with this usage doesn't mean it is correct. :)

lib/google/gax/config.rb
blowmage: Ugh, I went to add tests for this and it actually didn't raise. But it does now. It should now match the behavior of `attr_acccessor`:

```ruby
attr_acccessor "some method" # NameError (invalid attribute name `some method')
config_attr "some method", nil, Numeric # NameError (invalid config name some method)
```

lib/google/gax/config.rb
blowmage: This PR is meant to be a drop-in replacement for #195, but one thing that using this DSL does that the previous Configuration implementation doesn't is allow us to use Ruby inheritance. This way, a default can be provided whenever an attribute is defined, and new configs can inherit from an existing config class. So, for example, I hope to end up with something like following if we go with this approach:

```ruby
class Google::Cloud::Vision::Configure
  extend Google::Gax::Config

  config_attr :timeout, 60, Numeric
end

class Google::Cloud::Vision::V1::ImageAnnotator::Client::Configure < Google::Cloud::Vision::Configure
  config_attr :host, "vision.googleapis.com", String
  config_attr :port, 443, Integer
  # timeout is already configured
end
```

But we still have the instance parent/derived interplay:

```ruby
parent_config = Google::Cloud::Vision::V1::ImageAnnotator::Client::Configure.new do |config|
  config.timeout = 90
end

derived_config = Google::Cloud::Vision::V1::ImageAnnotator::Client::Configure.new(parent_config) do |config|
  config.timeout = 90
end

parent_config.timeout #=> 90
derived_config.timeout #=> 120

derived_config.timeout = nil # reset to parent config
derived_config.timeout #=> 90
parent_config.timeout #=> 90

parent_config.timeout = nil # reset to default
parent_config.timeout #=> 60
derived_config.timeout #=> 60
```

lib/google/gax/config.rb
```diff
@@ -0,0 +1,117 @@
...
+        # create getter
+        define_method name do
+          return instance_variable_get name_ivar if instance_variable_defined? name_ivar
+
+          if instance_variable_defined? :@parent_config
+            parent = instance_variable_get :@parent_config
+            return parent.send name if parent&.respond_to? name
+
+            parent = instance_variable_get :@parent_config
+            return parent.send name if parent&.respond_to? name
```
dazuma: Duplicated lines.

dazuma (approved):
Logic looks good to me now.

lib/google/gax/config.rb
blowmage: Yep. Removed.

PR #200, opened 2019-05-24
Make headers on operations client injectable

samphippen:
For Google Ads we send additional headers for LRO reloads, specifically:

```
developer-token
login-customer-id
```

without these headers being injectable it's not possible for us to
complete LRO calls successfully. This patch exposes headers in the
constructor of the operations client so that we can pass these when we
construct our operations client objects.

### DIFF
--- a/lib/google/longrunning/operations_client.rb
+++ b/lib/google/longrunning/operations_client.rb
@@ -114,7 +114,8 @@ module Google
           client_config: {},
           timeout: DEFAULT_TIMEOUT,
           lib_name: nil,
-          lib_version: ""
+          lib_version: "",
+          _headers: {}
         # These require statements are intentionally placed here to initialize
         # the gRPC module only when it's required.
         # See https://github.com/googleapis/toolkit/issues/446
@@ -147,7 +148,7 @@ module Google
         google_api_client << " grpc/#{GRPC::VERSION}"
         google_api_client.freeze
 
-        headers = { :"x-goog-api-client" => google_api_client }
+        headers = { :"x-goog-api-client" => google_api_client }.merge(_headers)
         client_config_file = Pathname.new(__dir__).join(
           "operations_client_config.json"
         )


dazuma:
@samphippen Just to be clear, this is for the *current* gapic generated clients, not the new generator, right? Do you have a similar requirement for the new generator?

samphippen:
@dazuma yes, we would need this in the new generator, but we need this for this generator right now for our LROs to work.

lib/google/longrunning/operations_client.rb
```diff
@@ -114,7 +114,8 @@ def initialize \
           client_config: {},
           timeout: DEFAULT_TIMEOUT,
           lib_name: nil,
-          lib_version: ""
+          lib_version: "",
+          _headers: {}
```
blowmage: I think `metadata` makes more sense than `_headers`.

```suggestion
          metadata: {}
```

dazuma: I can understand the naming. It's specifically injecting http headers. Maybe `extra_headers`?

blowmage: I think of it as the inverse. Users can provide common headers/metadata to use on every RPC call when creating the GAPIC client, and the GAPIC client adds the `x-goog-api-client` header to the headers/metadata that users provide.

Additionally, users can provide metadata on the individual RPC calls, and the common medatadata set on the client is merged with the per-RPC metadata.

samphippen:
I dug in to this further, and I think there's a bug with per-rpc headers: https://github.com/googleapis/gax-ruby/blob/1.x/lib/google/gax/settings.rb#L128

This line of code
```
metadata = (metadata.dup if metadata) || {}
```

is attempting to use the `metadata` `attr_accessor` on the `Settings` instance. However, in ruby, creating a local variable implicitly binds it to nil in the body of the block:

```
p foo
foo = 3
```

will print `nil` for example. This local binding also beats methods of the same name. As such `metadata` on that line will always reference the local variable `metadata` which will be bound to `nil`, and as such this entire expression will always assign the empty hash from the `|| {}` part of the line.

blowmage:
Yeah, I can confirm that line is not working as expected. It looks like changing that line to `metadata = @metadata.dup || {}` would fix it.

lib/google/longrunning/operations_client.rb
```diff
@@ -147,7 +148,7 @@ def initialize \
         google_api_client << " grpc/#{GRPC::VERSION}"
         google_api_client.freeze
 
-        headers = { :"x-goog-api-client" => google_api_client }
+        headers = { :"x-goog-api-client" => google_api_client }.merge(_headers)
```
blowmage: This can also be written like so:

```suggestion
        metadata ||= {}
        metadata[ :"x-goog-api-client"] ||= google_api_client
```

blowmage:
Closing in favor of #205.

PR #202, opened 2019-05-28
Allow Signet::OAuth2::Client as credential arg

blowmage:
Google Cloud supports `Signet::OAuth2::Client` as a credentials argument now. The `Credentials` class was originally intended to be a factory for creating `Signet::OAuth2::Client` objects, so it makes sense to support these objects.

GAPIC Generator adds support for this class to the configuration in googleapis/gapic-generator-ruby#151.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -23,6 +23,7 @@ Metrics/MethodLength:
 Metrics/PerceivedComplexity:
   Exclude:
     - "lib/google/gax/config.rb"
+    - "lib/google/gax/grpc/stub.rb"
 
 Style/Documentation:
   Exclude:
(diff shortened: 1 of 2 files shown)


.rubocop.yml
```diff
@@ -23,6 +23,7 @@ Metrics/MethodLength:
 Metrics/PerceivedComplexity:
   Exclude:
     - "lib/google/gax/config.rb"
+    - "lib/google/gax/grpc/stub.rb"
```
dazuma: I'm going to submit an issue to clean up these exclusions. Ideally we won't be excluding "normal" files from our rubocop settings.

dazuma (approved):
Approved, with a caveat.

PR #205, opened 2019-05-29
Add OperationsClient metadata argument

blowmage:
This PR is an alternate version of #200 that adds an argument to specify GRPC headers to be sent with the `OperationsClient` calls.


quartzmo (approved):
LGTM. There is no need to preserve the statement `google_api_client.freeze`, is there?

PR #206, opened 2019-05-29
Refactor Gax::GRPC::Stub to a class

blowmage:
Implement calling the RPC by calling a method on the Stub object instead of creating an ApiCall object. This allows us to fix the code that calls APIs in Gax instead of in the generator.

Generator support added in googleapis/gapic-generator-ruby#152.

### DIFF
--- a/lib/google/gax/grpc/stub.rb
+++ b/lib/google/gax/grpc/stub.rb
@@ -39,13 +39,16 @@ module Google
       #
       # This class wraps the actual gRPC Stub object and it's RPC methods.
       #
+      # @!attribute [r] grpc_stub
+      #   @return [Object] The instance of the gRPC stub class (`grpc_stub_class`) constructor argument.
+      #
       class Stub
-        attr_reader :stub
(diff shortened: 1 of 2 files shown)


lib/google/gax/grpc/stub.rb
```diff
@@ -29,16 +29,21 @@
...
       ##
-      # Stub
-      # TODO: Add documentation
-      module Stub
+      # Gax gRPC Stub
+      #
+      # This class wraps the actual gRPC Stub object and it's RPC methods.
+      #
+      class Stub
+        attr_reader :stub
```
dazuma: `wrapped_stub`? Might be a bit confusing for a "stub" to have a "stub" field.

dazuma (approved):
I like this. Less boilerplate in the generated code is generally a win in my book.

lib/google/gax/grpc/stub.rb
blowmage: What about `grpc_stub`?

PR #208, opened 2019-05-30
Call GAPIC client using positional args

blowmage:
This PR changes the Operations calls to the Operations Client to use positional arguments instead of the mixed positional/named arguments. This is needed to work around Protobuf [adding the `#to_hash` method to the resource classes](https://github.com/protocolbuffers/protobuf/pull/6166).

The generator side of this change is in googleapis/gapic-generator-ruby#155.


blowmage:
@quartzmo Since you approved googleapis/gapic-generator-ruby#155 can you approve this we well?

quartzmo (approved):

PR #210, opened 2019-05-30
Add headers helpers

blowmage:
Moving this common code to Gax so the GAPIC client can do less.

Generator uses this change in googleapis/gapic-generator-ruby#157.

### DIFF
--- a/lib/google/gax/headers.rb
+++ b/lib/google/gax/headers.rb
@@ -27,18 +27,33 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-require "google/gax/api_call"
-require "google/gax/configuration"
-require "google/gax/errors"
-require "google/gax/headers"
-require "google/gax/operation"
-require "google/gax/paged_enumerable"
-require "google/gax/protobuf"
-require "google/gax/stream_input"
-require "google/gax/version"
+require "google/gax/operation/retry_policy"
+require "google/protobuf/well_known_types"
 
 module Google
-  # Gax defines Google API extensions
   module Gax
+    # A collection of common header values.
+    module Headers
+      ##
+      # @param ruby_version [String] The ruby version. Defaults to `RUBY_VERSION`.
+      # @param lib_name [String] The client library name.
+      # @param lib_version [String] The client library version.
+      # @param gapic_version [String] The GAPIC client version.
+      # @param gax_version [String] The Gax version. Defaults to `Google::Gax::VERSION`.
+      # @param grpc_version [String] The GRPC version. Defaults to `GRPC::VERSION`.
+      def self.x_goog_api_client ruby_version: nil, lib_name: nil, lib_version: nil,
+                                 gapic_version: nil, gax_version: nil, grpc_version: nil
+        ruby_version ||= ::RUBY_VERSION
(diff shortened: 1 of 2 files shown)


lib/google/gax/headers.rb
```diff
@@ -0,0 +1,59 @@
+# Copyright 2016, Google LLC
```
quartzmo: 2019

lib/google/gax/headers.rb
```diff
@@ -0,0 +1,59 @@
...
+      def self.x_goog_api_client ruby_version: nil, lib_name: nil, lib_version: nil,
+                                 gapic_version: nil, gax_version: nil, grpc_version: nil
+        ruby_version ||= ::RUBY_VERSION
+        gax_version  ||= ::Google::Gax::VERSION
+        grpc_version ||= ::GRPC::VERSION if defined? ::GRPC
+
+        x_goog_api_client_header = ["gl-ruby/#{ruby_version}"]
+        x_goog_api_client_header << "#{lib_name}/#{lib_version}" if lib_name
+        x_goog_api_client_header << "gapic/#{gapic_version}" if gapic_version
+        x_goog_api_client_header << "gax/#{gax_version}"
```
quartzmo: impossible for `gax_version` to be nil, right? since `Google::Gax::VERSION` is present in the library?

test/google/gax/headers/x_goog_api_client_test.rb
```diff
@@ -0,0 +1,55 @@
...
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+require "test_helper"
+require "grpc"
+
+describe Google::Gax::Headers, :x_goog_api_client do
+  focus
```
quartzmo: remove `focus` from these tests

test/google/gax/headers/x_goog_api_client_test.rb
```diff
@@ -0,0 +1,55 @@
...
+  it "with no arguments" do
+    header = Google::Gax::Headers.x_goog_api_client
+    _(header).must_equal "gl-ruby/#{RUBY_VERSION} gax/#{Google::Gax::VERSION} grpc/#{GRPC::VERSION}"
```
quartzmo: It looks like these assertions will fail to prevent a blank value in the expected string in case any of these constants are nil or blank, is that a problem?

test/google/gax/headers/x_goog_api_client_test.rb
blowmage: LOL! This file had changes but wasn't added to the commit. This is how I figured out not all the tests were being run. Fixed.

lib/google/gax/headers.rb
blowmage: Right.

lib/google/gax/headers.rb
blowmage: Fixed.

test/google/gax/headers/x_goog_api_client_test.rb
blowmage: I don't think so. I expect half of these arguments to not be used, but are there for completeness/convenience.

quartzmo (approved):

PR #212, opened 2019-06-03
Override retry options even if no retry codes were specified

randypiper:
Previously, if a call's original `CallSettings` didn't specify any retry codes, they couldn't be overridden per-call via `CallOptions`. This change fixes that by checking the merged `CallSettings` (instead of the original `CallSettings`) to determine if a call should be retryable.

### DIFF
--- a/lib/google/gax/api_callable.rb
+++ b/lib/google/gax/api_callable.rb
@@ -249,7 +249,7 @@ module Google
           params = params_extractor.call(request)
           this_settings = with_routing_header(this_settings, params)
         end
-        api_call = if settings.retry_codes?
+        api_call = if this_settings.retry_codes?
                      retryable(func, this_settings.retry_options,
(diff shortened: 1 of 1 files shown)


quartzmo:
I ran the google-cloud-datastore `ci` and `acceptance` tasks against this branch and both passed.

```sh
Using google-gax 1.6.2 from https://github.com/randypiper/gax-ruby.git (at 1.x@672f2f5)
```

blowmage (approved):
This looks correct to me. @dazuma?

quartzmo:
@randypiper Would it be possible to add unit test coverage for this fix to `spec/google/gax/api_callable_spec.rb`?

dazuma (approved):
Yes, this looks right to me too.

quartzmo (approved):
I can attempt to add a unit test in a follow-up PR.

PR #213, opened 2019-06-04
Disable gRPC service config lookup by default (1.x branch)

dazuma:
Set the `grpc.service_config_disable_resolution` option on grpc channels by default, in the 1.x branch. See https://github.com/googleapis/gapic-generator/issues/2778. We will probably need to do something similar in the master branch.

/cc @noahdietz

### DIFF
--- a/lib/google/gax/grpc.rb
+++ b/lib/google/gax/grpc.rb
@@ -122,11 +122,13 @@ module Google
                       interceptors: [])
         verify_params(channel, chan_creds, updater_proc)
         address = "#{service_path}:#{port}"
+        default_channel_args = {"grpc.service_config_disable_resolution" => 1}
         if channel
           yield(address, nil, channel_override: channel,
                               interceptors: interceptors)
         elsif chan_creds
-          yield(address, chan_creds, interceptors: interceptors)
+          yield(address, chan_creds, interceptors: interceptors,
+                                     channel_args: default_channel_args)
         else
           if updater_proc.nil?
             auth_creds = Google::Auth.get_application_default(scopes)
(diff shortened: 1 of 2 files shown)


dazuma:
Looks like I'll need to update the mocks in the tests...

blowmage:
@dazuma The other PR says: "focusing this work only on micro-generators" If so, do we need to make this change on 1.x?

dazuma:
@blowmage The following comment in the other issue then says: "On second thought, we want clients to be safe as early as possible so we should disable it here as well asap"

blowmage:
Wait, so you are telling me reading comprehension still a thing? man...

lib/google/gax/grpc.rb
```diff
@@ -122,19 +122,22 @@ def create_stub(service_path,
                       interceptors: [])
         verify_params(channel, chan_creds, updater_proc)
         address = "#{service_path}:#{port}"
+        default_channel_args = { 'grpc.service_config_disable_resolution' => 1 }
```
blowmage: We have a few Google Cloud gems that create `GRPC::Core::Channel` objects because they need to set other channel args. Those gems will not benefit from this change. I would prefer to see `channel_args` added as an argument to `Google::Gax. create_stub` _and_ the GAPIC client constructor so we can pass the arguments instead of creating and passing a channel object.

blowmage: Once we have those two changes, we can update the Google Cloud gems to pass `channel_args` instead of creating a channel object.

quartzmo (approved):
This PR doesn't cover clients that pass `channel`, but as long as those are handled separately, LGTM!

blowmage (approved):

PR #215, opened 2019-06-26
Support overriding service address and port via parameters

dazuma:
This is to support overriding the backend service address. (Currently it is _possible_ in the clients by subclassing the client class, but it still fails for the associated Longrunning client and there doesn't seem to be a recourse other than monkeypatching.) This will go into the 1.x branch. In the 2.0 branch, this should be controlled by configuration.


quartzmo (approved):

blowmage (approved):
LGTM assuming these args are also being added to other GAPIC clients.

dazuma:
Yes, I'm synth-ing these into the GAPIC clients, but we need to start with this one to make sure the LRO client follows suit.

PR #219, opened 2019-08-14
fix: Honor per-call timeout overrides from client configs

### LINKED ISSUE #218: Per-method timeouts from client config json are not honored
Client config JSON files generated by gapic-generator sometimes include a `timeout_millis` field that sets the preferred default timeout for each call. (This value is used in the non-retry case. See `Google::Gax.add_timeout_arg`.)

As far as I can tell, this value is not being honored. Instead, the default timeout set in the gapic-generated client class is used. (This value seems to be hard-coded at 30s.)

This may be causing problems for services like Tasks, which currently implements a server-side limit of 30s on the timeout, and thus results in intermittent failures of this precondition.

This should probably be fixed in `settings.rb` in `Google::Gax.construct_settings` by reading any `timeout_millis` from the `overriding_method` hash.

dazuma:
Fixes #218

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,5 +4,7 @@ inherit_from: .rubocop_todo.yml
 # when exploded exception style is used.
 Metrics/ClassLength:
   Max: 110
+Metrics/ModuleLength:
+  Max: 110
 Style/RaiseArgs:
   EnforcedStyle: compact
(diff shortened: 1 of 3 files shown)


blowmage:
Rubocop is failing...

blowmage (approved):

PR #225, opened 2019-10-29
Allow spaces in path template render bindings

blowmage:
This change allows spaces to be present in the bindings passed to `PathTemplate#render`.


blowmage:
This is related to googleapis/google-cloud-ruby#4291

dazuma (approved):

PR #228, opened 2019-11-11
PageEnumerable return Enumerator when not passing blocks

blowmage:


dazuma (approved):

dazuma:
google-gax is no longer used in current client versions. Closing all PRs.

